PR #16, opened 2023-07-19
Adding Usersnap integration

ajolipa:
### What this does
Added a script in `BaseLayout.astro` to load the Usersnap feedback widget.
image
image
 

### Notes and further work
I have not yet done much to style the widget (other than making the background the same blue as the menu -- I tried to make it the right font but it doesn't seem to be loading correctly) or tweak its settings on the Usersnap side (for example I assume we don't want the field for choosing an assignee to show up), so there's plenty more to tinker with. It also currently is in the base layout so should show up on every page, but we can do something fancier with that as well.


rsimon:
Perfect, thanks! Merging now.

PR #26, opened 2023-08-02
Policy querying in the client + (some) policy-aware UI components

rsimon:
## In this PR

This PR adds client-side plumbing to use the new backend RPCs for policy-querying in a convenient way. 

```
policy.get('<table-name>').has('<SELECT | INSERT | UPDATE | DELETE>');
```

This PR also makes the following UI elements policy-aware:
- __Create New Project__ button in the dashboard, by fetching the users organization policies on load, and then checking whether the user has __INSERT__ permissions on the `projects` table.
- __Edit Project Description__ button in the Projects Home page, by checking the project policies for __UPDATE__ permissions on the `projects` table.
- Document upload (button + drag and drop behavior) in the Project Home page, by checking the project policies for __INSERT__ permissions on the `documents` table.

CC-ing @lwjameson for a quick sanity check on the approach.

__UPDATE:__

The latest commit also adds policy-awareness to the initial empty state of the dashboard. There is extra "onboarding" text, which also takes into account whether the user already has invitations waiting. Texts localized in EN and DE.

 
Bildschirmfoto 2023-08-02 um 14 24 02


ajolipa (approved):
I like the new empty dashboard onboarding language.

PR #32, opened 2023-08-11
'All Projects' tab for non-admin users

rsimon:
## In this PR

This is another PR in response to beta user feedback. People were confused when projects they joined didn't show up in the dashboard. (Because they'd only show up under "Shared with Me", and they might have been in "My Projects".)

This PR adds the "All Projects" back for non-admin users. If you are not an admin, the "All Projects" will simply display "Mine" + "Shared with Me" (rather than all projects actually returned from the backend, which may include additional projects you don't actually have access to.)

Here's a comparison of a what an Org Professor user will see vs. an Org Admin:

Bildschirmfoto 2023-08-11 um 10 26 01

Bildschirmfoto 2023-08-11 um 10 27 58

### DIFF
--- a/src/apps/dashboard-projects/Header/Header.css
+++ b/src/apps/dashboard-projects/Header/Header.css
@@ -65,7 +65,7 @@
 }
 
 .dashboard-header-bottom li + li {
-  margin-left: 10px;
+  margin-left: 18px;
 }
 
 .dashboard-header-bottom li button {
--- a/src/apps/dashboard-projects/Header/Header.tsx
+++ b/src/apps/dashboard-projects/Header/Header.tsx
@@ -99,18 +99,16 @@ export const Header = (props: HeaderProps) => {
 
       <section className="dashboard-header-bottom">
         <ul className="dashboard-header-tabs">
-          {props.me.isOrgAdmin && (
-            <li
-              className={filter === ProjectFilter.ALL ? 'active' : undefined}
-              onClick={() => onChangeFilter(ProjectFilter.ALL)}>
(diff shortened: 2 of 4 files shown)


rsimon:
A note to @lwjameson: I ended up not using the new  `get_my_projects` RPC after all. That would have introduced yet another DB roundtrip. 

The query that populates the dashboard has become quite complex. Here's what I'm doing at the moment.

- Query the `projects` table, JOINing
  - creator profile
  - contexts
  - layers, JOINing
    - documents
  - project_groups 
- Pull **all** the group IDs from **all** the projects in the response
- Run a follow-up query to get all the group members for all the group members for all the groups
  - The follow up query JOINs the user profiles into the response
- Process the results to weave the user profiles back into the project groups

The reason for the complexity is that I'm showing the following data in the dashboard UI:
- User avatars for project members
- Icons for number of text and image documents

Note that creator profile might not actually be needed any more, since I've added the group member follow-up. Also, contexts are not used in the dashboard. But I kept them in for consistency with the project details query for a single project (and it might come handy in the future, if we want to show an icon for number of assignments).

For info, the code for the whole query is here. Main query:
https://github.com/recogito/recogito-client/blob/main/src/backend/helpers/projectHelpers.ts#L89-L156 

`getGroupMembers` follow-up:
https://github.com/recogito/recogito-client/blob/main/src/backend/helpers/projectHelpers.ts#L229-L245

In the long run we should maybe think about moving the whole complexity into a single server function.

ajolipa (approved):

PR #35, opened 2023-08-18
implemented limit of five on displayed avatars

ajolipa:
### In this PR
This PR implements a limit (currently set to five but can be changed pretty easily) to the number of avatars displayed in the top right of a document, putting the remaining ones in a dropdown.
image

### Notes and further work needed
- Currently only implemented in desktop text annotations, not images. Once completed it will need to be added to image annotation component as well.
- Max height of dropdown not implemented yet
- Currently dropdown only shows the users that were left off the list, but also it shows names and not just avatars, which is a little weird because that means you can only see the names of people other than the first five.... Could be changed so that the dropdown shows all users, or maybe have a tooltip over the existing avatars that displays the name? Open to suggestions.
- "+n more" language needs to be localized; currently hard-coded in English

I would recommend *not merging* this PR until some of these tweaks are made, but I'm putting it here now for review/suggestions and to generate the netlify preview.

### DIFF
--- a/src/apps/annotation-text/TextAnnotationDesktop.tsx
+++ b/src/apps/annotation-text/TextAnnotationDesktop.tsx
@@ -16,6 +16,7 @@ import {
 import './TextAnnotationDesktop.css';
 
 import '@recogito/react-text-annotator/dist/react-text-annotator.css';
+import { PresenceViewMoreButton } from '@components/Presence/PresenceViewMoreButton';
 
 const SUPABASE = import.meta.env.PUBLIC_SUPABASE;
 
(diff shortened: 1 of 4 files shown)


rsimon:
I love it! 

But, yes to all your points about remaining work. It definitely needs scrolling (e.g. you could limit the `max-height` to 50vh or something?). And it should show all users, not just those that were left off, as you say. (I can handle the localizing, that's the easy part :-) 

Two more comments: 
- Since there's nothing "to do" with the avatars in the list, they shouldn't be (or look) clickable. Right now there's a hover effect which I think is confusing since nothing will happen on click. I think you should use the [Radix popover component](https://www.radix-ui.com/primitives/docs/components/popover) instead of the [dropdown menu](https://www.radix-ui.com/primitives/docs/components/popover), and then inside the popover just an ordinary `<ul>` list.
- About integrating with the image annotation desktop: I would definitely make the whole component part of the `PresenceStack` component. Semantically it's supposed to be one component (5 avatars + dropdown list) anyway. And once you make it part of the `PresenceStack`, there's no need to have (redundant) extra code in the image and text annotation views.

ajolipa:
Implement the above suggestions -- let me know how it looks now! Changes from previous commit:
- Moved `view more` button into the `PresenceStack` component
- Changed from using a dropdown menu to a popover for the content
- Popover now displays all active users, not just the overflow ones
- Popover content has a max height of 50vh and will scroll if there's overflow

Further work needed:
- Localization of the button text is still not done

PR #38, opened 2023-08-21
Password Reset

rsimon:
## In this PR

This PR adds UI for password reset. The UI is located at `/reset-password`. Supabase password-reset emails should point to this path. The UI will auto-forward to the localized version (e.g. `/de/reset-password`).

Bildschirmfoto 2023-08-21 um 14 14 01

Bildschirmfoto 2023-08-21 um 14 14 13


rsimon:
__Update:__ added UI for requesting a password-reset link.

Bildschirmfoto 2023-08-21 um 15 28 27

ajolipa (approved):
UI looks great. When I tried to actually use it I didn't receive an email, and when I tried to send one directly from the Supabase dashboard I got this error -- maybe just because a lot of requests have been sent as you're working on this feature? 
image
So that seems like an issue on the Supabase side, but is there any way for the client to tell whether the email was actually sent, and display some sort of error message if Supabase returned an error?

rsimon:
I made a few additional commits. You should see an error displayed now. I'm also getting the 'rate limit' exceeded error, and have been for the past 30mins or so. Not sure what the settings are on Supabase for lifting that limit again. But it seems weird that you are getting them, too - for a different email address, and from another computer. So possibly some kind of global limit for no. of emails that Supabase will send out for a free account?

PR #39, opened 2023-08-22
Basic TEI Support

rsimon:
## In this PR

This PR adds a first working version of TEI support. There are currently two known issues:

- __There is a performance problem with very large TEI files.__ Most TEI files (hopefully!) will not fall into this category. [The sample file here](https://github.com/gregorycrane/Homerica/blob/master/chs-odyssey.xml), for example, is a >750kB file which works just fine. But larger files will currently result in sluggish UI behavior. The problem is fixable, and caused by a bit of unfortunate design of the architecture on my part. (The state store still operates with text offset selectors internally, and they are converted to/from TEI XPath selectors on the "edges" of the API, which results in a lot of conversions. Because conversions involve DOM ranges, this gets slower as the DOM gets bigger. The solution is to redesign the architecture, so that the state store uses XPath selectors natively, which will massively reduce the number of conversions happening. I originally tried to avoid rebuilding the store like this, since it might introduce extra redundancy in the code, but we'll see.)

- __Some annotations look like they are multiple (overlapping) annotations.__ The screenshot below shows the effect - what's depicted is really only __one__ annotation. I still need to understand the effect fully. But here's what I think is going on:
  - The annotation selector is correct, i.e. this is really just one annotation, as it should be, with correct XPath start and end points.
  - Rendering is based on a DOM Range that is created between the annotation's start and end point. Likewise, this seems to work correctly.
  - From the DOM range, I'm retrieving the set of rectangles to draw using `range.getClientRects()`. It looks like this method returns multiple overlapping rectangles when a nested markup structure is involved. To fix this, it's likely going to be necessary to take all rectangle shapes, and compute a set of union rectangles.

Bildschirmfoto 2023-08-22 um 12 19 59


rsimon:
__An update on the performance problem:__ looks like there was an additional bug involved that caused lots of exceptions, which would then cause a slowdown. After fixing this, performance is much better. 

While the TEI annotation plugin design is still not ideal, I think we can leave it like that for the time being. (I tested with a 2MB TEI file and it worked quite well.)

PR #41, opened 2023-08-24
Feature: Tagging

rsimon:
## In this PR

This PR adds tagging functionality to the annotation interface. This is more of first-wild-sprint implementation. Some things don't work properly yet, and I'm not quite happy with the design either. But in terms of functionality, it get's the job done ;-) Some screenshots below. Things that definitely need more work:

- The row of tags should always show the 'Add tag' field, until there is not enough space left. After this, the 'Add tag' field should be replace with the '+' button that opens the tag input field in a popover. Right now, the 'Add Tag' field is hidden after 3 tags exist, regardless of space.
- The tag input field popover alignment should be "end", not "start".
- The design with the tag section at the top IMO works well for the popup. But it feels like it takes up way too much space in the annotation list. We may need different designs between popup and list. (Maybe just an un-styled inline list at the bottom? With a small + icon that triggers the tag input popover?)
- The 'More tags' popover has a fixed width. It should auto-scale (between a min and a max width)
- Not sure I like the tooltip-style delete button...

Bildschirmfoto 2023-08-24 um 17 10 47

Bildschirmfoto 2023-08-24 um 17 11 18

Bildschirmfoto 2023-08-24 um 17 11 35

Bildschirmfoto 2023-08-24 um 17 11 51

Bildschirmfoto 2023-08-24 um 17 12 02


ajolipa (approved):
Looks good to me. Agreed re: taking up too much space in the list, but that's not a big deal for now.

PR #49, opened 2023-09-07
Attempt to fix issue #127

rsimon:
## In this PR

I have diagnosed issue [#127](https://github.com/performant-software/vico/issues/127) and it seems like a weird timing issue.

- When annotations are initially loaded, each annotation is parsed, and start- and end-offsets are converted to a DOM Range object.
- Based on this object, a 2D index is built which stores all the rectangle shapes for each range, in each annotation.
- For Canvas rendering, the 2D index is queried for the annotations that intersect the current viewport.

The problem is this: on initial load, the function that returns the revived annotations (i.e. those that had their DOM Range computed based on start/end) frequently returns zero-length ranges. When refreshing the page, this doesn't happen any more.

It looks like a) Ranges are returned before they are fully computed; and b) the process runs faster once it has happened once. It's pretty hard to debug, because the effect is no longer reproducible after a refresh.

I have now introduced a retry mechanism: if collapsed ranges are received, the renderer waits 250ms and tries again. This happens at most 5 times, after which whatever is there is rendered to the screen. In addition, I added debug logging to further understand the issue.

I'm issuing this PR now, so I can test with the Netlify PR preview.


rsimon:
__Update:__ looks like it's a different issue, [described here](https://github.com/performant-software/vico/issues/127).

PR #64, opened 2023-10-13
https://github.com/performant-software/cove-recogito/issues/1

lwjameson:
This implements reading branding and authentication methods from the config file.

This implements a new landing page for login:

image

There is a minimal default config which is used for development.  This config is replaced with the organizations config at build time.

### DIFF
--- a/src/apps/auth-login/Login.tsx
+++ b/src/apps/auth-login/Login.tsx
@@ -1,17 +1,15 @@
-import { useEffect, useState } from "react";
-import type { Session } from "@supabase/supabase-js";
-import { supabase } from "@backend/supabaseBrowserClient";
-import { isLoggedIn } from "@backend/auth";
-import type { LoginMethod, Translations } from "src/Types";
-import { StateChecking, StateLoginForm, StateMagicLink } from "./states";
-import { LoginMethodSelector } from "@apps/auth-login/LoginMethodSelector";
+import { useEffect, useState } from 'react';
+import type { Session } from '@supabase/supabase-js';
+import { supabase } from '@backend/supabaseBrowserClient';
+import { isLoggedIn } from '@backend/auth';
+import type { LoginMethod, Translations } from 'src/Types';
+import { StateChecking, StateLoginForm, StateMagicLink } from './states';
+import { LoginMethodSelector } from '@apps/auth-login/LoginMethodSelector';
 
-import "./Login.css";
-
-const SSO_DOMAIN = import.meta.env.PUBLIC_SSO_DOMAIN;
+import './Login.css';
 
 const setCookies = (session: Session | null) => {
-  if (!session) throw "SIGNED_IN event without session - should never happen";
+  if (!session) throw 'SIGNED_IN event without session - should never happen';
 
   const maxAge = 100 * 365 * 24 * 60 * 60; // 100 years, never expires
   document.cookie = `sb-access-token=${session?.access_token}; path=/; max-age=${maxAge}; SameSite=Lax; secure`;
@@ -38,9 +36,9 @@ export const Login = (props: {
 
   useEffect(() => {
     supabase.auth.onAuthStateChange((event, session) => {
-      if (event === "SIGNED_OUT") {
+      if (event === 'SIGNED_OUT') {
         clearCookies();
-      } else if (event === "SIGNED_IN" || event === "TOKEN_REFRESHED") {
+      } else if (event === 'SIGNED_IN' || event === 'TOKEN_REFRESHED') {
         setCookies(session);
         window.location.href = `/${props.i18n.lang}/projects`;
       }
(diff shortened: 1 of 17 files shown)


src/config.json
```diff
@@ -0,0 +1,29 @@
+{
+  "_comment": "This is a default configuration file that can be replaced a build time",
```
rsimon, line 2: "replaced a build time" typo ;-)

src/layouts/login/LoginLayout.astro
```diff
@@ -0,0 +1,122 @@
+---
+import BaseLayout from "../BaseLayout.astro";
+import type {
```
rsimon, line 3: Most of these imports seem to be unused. Specifically all the imports from `Types`, and the `LoginMethodSelector`. Also, Cookies aren't accessed, so this line can also be deleted.

src/layouts/login/LoginLayout.astro
```diff
@@ -0,0 +1,122 @@
...
+  ExtendedProjectData,
+  MyProfile,
+  LoginMethod,
+} from "../../Types";
+import config from "../../config.json";
+import { LoginMethodSelector } from "@apps/auth-login/LoginMethodSelector";
+
+export interface Props {
+  title: string;
+  i18n: Node;
```
rsimon, line 14: I'm getting an error in `sign-in.astro` because of this. i18n isn't a Node, but of Type `Translations`.

src/apps/auth-login/LoginMethodSelector/LoginMethodSelector.tsx
```diff
@@ -0,0 +1,73 @@
+import { ReactNode, forwardRef, useState } from "react";
```
rsimon, line 1: There are some unused imports here: `useState`, `supabase` and `updateUserProjectGroup`.

src/apps/auth-login/Login.tsx
```diff
@@ -1,85 +1,136 @@
...
+import { supabase } from "@backend/supabaseBrowserClient";
+import { isLoggedIn } from "@backend/auth";
+import type { LoginMethod, Translations } from "src/Types";
+import { StateChecking, StateLoginForm, StateMagicLink } from "./states";
+import { LoginMethodSelector } from "@apps/auth-login/LoginMethodSelector";
 
-import './Login.css';
+import "./Login.css";
 
 const SSO_DOMAIN = import.meta.env.PUBLIC_SSO_DOMAIN;
```
rsimon, line 11: The SSO_DOMAIN env variable is now no longer used, I believe?

src/apps/auth-login/Login.tsx
```diff
@@ -1,85 +1,136 @@
...
-      domain: SSO_DOMAIN
-    }).then(({ data, error }) => {
-      if (data?.url) {
-        window.location.href= data.url;
-      } else {
-        console.error(error);
-      }
-    });
-  }
+  const signInWithSSO = (domain: string) => {
```
rsimon, line 61: I think this might need minimal error handling. When I tried locally this (naturally) failed (because I had not domain configured). 

Arguably, in a correctly configured systems this shouldn't normally fail. But you never know. Should we attach at least some kind of "Something went wrong" UI feedback to the error handler?

rsimon (changes requested):
Hi @lwjameson, thanks for this - Login form revisions look good! 

I have a few comments on cleanup-/housekeeping-type stuff, but nothing important, really: 

- code-style (quotes) were set to a different strategy in your IDE. If it's quick to fix, it would be great to stick with the original style
- a few stray unused imports here and there

My only (more general) remark: the branding now doesn't include the possibility for localization. We may want to think about this further down the line (e.g. providing labels for different languages right in the config file?). But, a) certainly not crucial in this round and, b) probably something we'll solve implicitly once we address the issue of "label-branding" more generally. (= the use case of, e.g., replacing "Projects" with "Anthologies").

src/apps/auth-login/Login.tsx
lwjameson: I have removed the variable because the config will be required.

src/apps/auth-login/Login.tsx
lwjameson: Makes sense. Error handling added

src/apps/auth-login/LoginMethodSelector/LoginMethodSelector.tsx
lwjameson: Fixed

src/apps/auth-login/Login.tsx
lwjameson: Actually I have not done this yet. I will come back to it but need this in the develop branch quickly

src/layouts/login/LoginLayout.astro
lwjameson: Fixed

rsimon (approved):
Looks all good now, thanks!

PR #71, opened 2023-10-24
Editable Assignments

rsimon:
## In this PR

This PR adds functionality to __edit an existing assignment__ (see https://github.com/performant-software/vico/issues/147, https://github.com/performant-software/vico/issues/154).

- The actions dropdown on the Assignment card now has an extra 'Edit Assignment' option. (Note that the actions menu is only available to admin level users)
- Selecting the 'Edit Assignment' menu option will bring up the same editing wizard that is also shown when creating a new assignment, but with all tabs pre-filled.
- It is possible to change all aspects of the Assignment:
   - Name
   - Documents
   - Team
   - Readme
- UI labels in the wizard are changed accordingly ('Save Assignment' instead of 'Create Assignment'), and localized labels were added
- When saving an existing assignment, the wizard performs different steps. Instead of creating, the following happens:
   - Name and description are checked for changes. The `context` record is updated accordingly, if needed.
   - Documents and team are checked for additions and removals
   - Document `layers` are created and archived as needed, if documents were added to the assignment. Users are added to the newly created document layers.
   - If there were additions or removals to the team, the `Layer Students` groups on existing documents are updated accordingly.

### DIFF
--- a/src/apps/project-assignments/Grid/AssignmentsGrid.tsx
+++ b/src/apps/project-assignments/Grid/AssignmentsGrid.tsx
@@ -11,8 +11,9 @@ interface AssignmentsGridProps {
 
   project: ExtendedProjectData;
 
-  // Just temporary, for hacking/testing
   assignments: Context[];
+
+  onEditAssignment(assignment: Context): void;
 
   onDeleteAssignment(assignment: Context): void;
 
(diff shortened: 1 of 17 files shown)


SteveMarvin:
This is working for me. It does say "Update Assignment" instead of the "Save assignment" indicated above, but that seems like a solid change.

rsimon:
Perfect, thanks!

lwjameson (approved):
Looks good. Approved for pull.

PR #77, opened 2023-10-31
Added PDF options to file upload dropdown

lwjameson:


rsimon (approved):
Looks good - and flawless German localization, too! ;-)

PR #79, opened 2023-11-03
Rich text editor

lwjameson:
This PR adds a rich text editor for bodies.


package.json
```diff
@@ -15,6 +15,7 @@
     "@astrojs/netlify": "^2.2.3",
     "@astrojs/node": "^5.3.3",
     "@astrojs/react": "^2.2.0",
+    "@lexical/react": "^0.12.2",
```
rsimon, line 18: I guess the Lexical dependencies can be removed, right?

package.json
```diff
@@ -38,12 +39,15 @@
     "confetti-explosion-react": "^0.1.2",
     "crypto-js": "^4.1.1",
     "formik": "^2.4.1",
+    "lexical": "^0.12.2",
```
rsimon, line 42: Same here?

src/components/Annotation/Card/BaseCard.tsx
```diff
@@ -9,64 +9,69 @@ import { TagsWidget } from '../TagsWidget';
...
 
 import './BaseCard.css';
+import type { Translations } from 'src/Types';
 
 type BaseCardProps = CardProps & {
-
   comment(props: CommentProps): ReactNode;
 
-}
+  i18n: Translations;
```
rsimon, line 17: Did your IDE complain about this. Shouldn't be necessary, since translations are already included in the `CardProps` base type.

src/components/Annotation/Comment/EditableComment.tsx
```diff
@@ -27,13 +25,19 @@ export const EditableComment = (props: EditableCommentProps) => {
...
   const textarea = useRef<HTMLTextAreaElement>(null);
 
-  const [value, setValue] = useState(comment.value);
+  const [value, setValue] = useState<string | undefined>(
+    // Hack for now
+    comment.value.length > 0 && comment.value.charAt(0) === '{'
+      ? JSON.parse(comment.value)
+      : comment.value
+  );
+  const [renderType, setRenderType] = useState<'text' | 'quill'>('text');
```
rsimon, line 34: BTW, I think the cleanest way to handle this is to make use of W3C Web Anno's own terminology. Quill vs. plaintext would essentially represent different content types, right? A body can have a `format` field. Formats _should_ comply to MIME types. But aren't strictly required to, so we could make up our own. Or we could simple use `text/plain` vs `application/json`.

src/components/Annotation/Comment/EditableComment.tsx
```diff
@@ -46,46 +50,78 @@ export const EditableComment = (props: EditableCommentProps) => {
...
+    }
+  }, []);
+
   const onSaveChange = (evt: React.FormEvent) => {
     evt.preventDefault();
 
     store.updateBody(comment, {
       ...comment,
-      value: textarea.current!.value
+      render_type: renderType,
```
rsimon, line 68: Yep, I think this should simply be replaced by `format` and then either with a custom value (`plaintext` vs. `quill` or `text/plain` vs. `application/json`.

src/components/Annotation/ReplyForm/ReplyForm.tsx
```diff
@@ -45,67 +47,58 @@ export const ReplyForm = (props: ReplyFormProps) => {
...
           id: me.id,
           name: me.name,
-          avatar: me.avatar
+          avatar: me.avatar,
         },
         created: new Date(),
         purpose: 'commenting',
-        value: textarea.current.value
+        value: JSON.stringify(value),
+        render_type: 'quill',
```
rsimon, line 67: Hm, this feels like a bit of a mixed approach. If we implement it like this, that means we'd never have a plaintext mode anymore. (Every reply would be hard-wired to rich text, if I understand correctly.)

I wonder what the purpose then would be for even keeping the dual support (editable text area and rich text). IMO we should decide between two approaches:
- Allow both rich text and plaintext. In this case, the ReplyForm would have to support both modes, too.
- Decide to go for Rich Text only. In this case, there would be no need for switching between components. The only thing we'd need to do is to check if the body value is plaintext or JSON, and crosswalk from plaintext to Quill format, to maintain backwards compatibility with older annotations.

src/i18n/de/annotation-image.json
```diff
@@ -1,3 +1,9 @@
 {
-  "Anonymous": "Anonym" 
+  "Anonymous": "Anonym",
```
rsimon, line 2: Wait, there's actually an `annotation-common.json` file that get's included in both (text and image) annotation label sets! I.e. this should better go in there. No need for duplication.

rsimon (changes requested):
Thanks for this! I left a few comments. I guess the key question is whether we even want to maintain dual mode (i.e. allow both plaintext AND rich text). If so, the ReplyField probably needs tweaking - see my comment below.

Also: are those image resources in the public folder still needed?

package.json
lwjameson: Done

package.json
lwjameson: Done

src/components/Annotation/Comment/EditableComment.tsx
lwjameson: I guess my reasoning here is that while quill uses a JSON object, so do other RTE frameworks.  If we replace or have a different option for the RTE than specifying JSON will not be enough to distinguish one framework from another.

src/components/Annotation/Comment/EditableComment.tsx
lwjameson: So you are saying replace render_type with format? And then defining a new format for quill? Where would I do that?

src/components/Annotation/ReplyForm/ReplyForm.tsx
lwjameson: I considered an inline upgrade to rich text but was not sure.  If you agree I can probably do an inline translation of all text on render and a conversion on edit.

src/i18n/de/annotation-image.json
lwjameson: Had no idea ;-) Moved to common.

src/components/Annotation/Comment/EditableComment.tsx
rsimon: Agreed about `application/json` not being enough. How about the following: 
- We use `text/plain` for the text comments, i.e. stick to the standard in this case (and potentially have interop with other tools when annotations are exported)
- We use `quill` as the format for the Quill comments?

src/components/Annotation/Comment/EditableComment.tsx
rsimon: My thought would be, simply, to really just replace `render_type` with `format`. No other change needed. `format` is already a defined field in the spec. And then we'd just have `text/plain` and `quill` as the two possible values of `format`. Would be a minimal change, but has the advantage of making annotations fully spec compliant, if they are plaintext; and still pretty much spec-compliant for Quill-annotations. ("Pretty much" in the sense that if we export the annotations, other WebAnno clients would be able to look up the `format` field, and decide that they can't just render it as plaintext.)

src/components/Annotation/ReplyForm/ReplyForm.tsx
rsimon: I think the question is whether we want the ReplyField to support plaintext at all. If so, we'd need another switching mechanism. Not sure how that would work though. Would have to be a config argument, or env variable passed from Astro. Essentially to tell the client "render a plaintext reply field by default" vs. "render the RTE by default". 

Really not sure what to do here, and no strong opinions. Some thoughts: (also CC-ing @jamiefolsom.)
- I believe it's really a matter of deciding whether we want to be able to switch the RTE feature on and off for specific clients in the future, or just give it to anyone. (My gut feeling is that we should just make it the default. Definitely keeps the code simpler)
- Did Bonn say they want rich text, too? If so, that might be our answer.

Caveat: Quill introduces a bit of a difficulty for the use case of __exporting annotations from Recogito__. When exporting to CSV (or W3C Web Anno) there's no simple way for us to produce readable output. Therefore, I wonder: 

- is there a way for us to cross-walk Quill content to plaintext on the server, during the export?
- if not, should we store two sets of values in the annotation? The Quill/JSON value __and__ and a plaintext version? (I'm assuming there is an easy way to pull an unformatted copy of the text from Quill - which may or may not be the case...)

src/components/Annotation/Comment/EditableComment.tsx
lwjameson: Sounds good.  I suggest we use application/quill which seems more in line other format options.

src/components/Annotation/Comment/EditableComment.tsx
lwjameson: I will replace with format.

src/components/Annotation/ReplyForm/ReplyForm.tsx
lwjameson: Yes.  Quill has a [getText](https://quilljs.com/docs/api/#gettext) method which converts the quill content to text, omitting non-text content.  You would just instantiate quill in your CSV or TEI export method and convert all of the quill bodies.

src/components/Annotation/Comment/EditableComment.tsx
lwjameson: So I have replace with format but that is not be persisted to the DB.  Can you take a look at that?

rsimon: Oh - good point. Come to think of it: I don't think `render_type` ever got persisted, either right? It looks like the decision in the client is based by inspecting the value, and checking if the first char is a `{`. 

Here's my suggestion:
- Good news is: the DB `bodies` table already have a `format` field. (We added that way back in the beginning but never used it.) The permitted values currently are `TextPlain` and `TextHtml`. Can you change that to `text/plain` and `text/quill`?
- Then I'll make sure that `format` actually gets stored. (This is a small fix in the [Annotorious/Supabase storage plugin](https://github.com/recogito/annotorious-supabase).)

src/components/Annotation/ReplyForm/ReplyForm.tsx
rsimon: It looks like quill requires access to the `document`, i.e. it won't work on the server. The Recogito data exports are just API endpoints, which means we can't crosswalk easily.

Unless we want to store two copies of the comment (quill and plaintext), we have two options:

- Build our own crosswalk, by figuring out how the Quill JSON format works and then stripping/concatenating the tokens ourselves
- Adding a client-side crosswalk to the text exporter.
  - The user would click the download button
  - There would be JS on the client that fetches the annotations, instantiates Quill, runs the conversion client side
  - After the conversion is done, there would be a [fake download link](https://stackoverflow.com/questions/19721439/download-json-object-as-a-file-from-browser) which would provide the same experience as an actual download from an API endpoint.

However: while technically possible, I'd rather not use the client-side hack! One of the first things that people wanted to do in old Recogito was to harvest data from the APIs. Once the exports rely on browser JS to provide the right data, we are preventing any kind of script interaction.

Therefore: I think that storing both versions or 'reverse-engineering' our own Quill crosswalk might be our only options.

src/components/Annotation/Comment/EditableComment.tsx
rsimon: __Update:__ I now made the changes to the Annotorious Supabase plugin. (See [here](https://github.com/recogito/annotorious-supabase/commit/1f1b683d4d28ce131c5cce2421d871a5ab15b2f8).) I updated the version in the `package.json`. I.e. if you run `git pul`, `npm install`, the format field should get persisted.

PR #80, opened 2023-11-06
TEI Export

rsimon:
## In this PR

This PR adds functionality to re-export an uploaded TEI document. Annotations created on the TEI document will be merged with the original TEI markup, as a `<standOff>` element in the `<teiHeader>`. (See https://github.com/performant-software/vico/issues/200 for a markup example.)

- Documents can be exported either at the __project level__ (default context), or at the __assignment level__.
   - Project-level exports are only accessible to Project Admin users, and include annotations from all assignment layers.
   - Assignment-level exports are accessible to every user with access to the assignment, and include annotations only from this assignment layer(s)
- The user can choose between exporting only public annotations, or having their own private annotations included in the export. 

Bildschirmfoto 2023-11-06 um 11 45 11


SteveMarvin:
I'm going to assume that this is meant to export the entire document WITH annotations, rather than a TEI file that only contains the annotations. With that in mind, APPROVED. ;-)

PR #82, opened 2023-11-13
Document Notes

rsimon:
## In this PR 

This PR adds __a beta version of the Document Notes__ feature. Documents notes work like annotations, but pertain to the document as a whole. They are not connected to a selected text or image shape, and (in terms of the W3C Web Annotation model) have no selector.

Notes are listed in the right sidebar, in a separate panel, and have a different styling (post-it-note yellow), so that users can differentiate them from annotations.

- In the backend, notes are stored just like ordinary annotations, along with a target and bodies. However, the `value` of the target is always `null`. 
- Notes can be private or public 
- Document notes are synchronized live like annotations, using the Supabase Realtime channel:
   -  using Postgres updates: INSERT on the targets table, INSERT and UPDATE on the bodies table
   - using the broadcast channel for events that cannot be handled via Postgres updates due to RLS: archiving a note, archiving a comment body, making a private note public
- The Notes sidebar lists notes in chronological order by default (newest first). Sort order can be reversed by the user. 

### Screenshot

Bildschirmfoto 2023-11-13 um 14 12 01


lwjameson (approved):
Looks good!

SteveMarvin:
This looks good!

PR #84, opened 2023-11-14
Fix for invites

lwjameson:
Invites were not using the RPC and started to fail once the policies were re-enabled on the invites table


rsimon (approved):
Looks good, thanks! 👍

PR #88, opened 2023-11-20
Consistent Color Coding in Multi-User Session

rsimon:
## In this PR

This PR changes the way color coding works. Before, colors were assigned randomly, based on the (arbitrary) order by which the renderer encountered new annotations. With this PR, a full color legend is built for __all__ annotations in the document, and kept in sync when annotations are added, deleted or changed.

The color coding is determined in a predictable way (e.g.: "all tags in the document, sorted by the time they were first used") This means different users will get the same color key, regardless of when they join the session.

There were also a few additional tweaks:
- renaming/refactoring to account for recent changes in the Annotorious API
- slightly increased opacity of the text highlights
- regression bugfix that caused selected text annotations to appear with the same opacity as unselected ones, instead of a higher opacity


lwjameson:
@rsimon Hi Rainer.  When I run this locally, the annotations are always immediately archived so do not appear.  Does this have anything to do with PR #90 ?

rsimon:
Can you try creating annotations with the sidebar open? If that works - yes, the behavior you're seeing is unrelated to this PR, and that bug was fixed in PR 90.

rsimon:
I merged the fix from PR #90 into this branch. This should now be testable.

lwjameson (approved):

PR #93, opened 2023-11-22
Prevent Zombie Annotations on page unload

rsimon:
## In this PR

We have taken various measures to prevent the creation of "zombie annotations" - annotations that have no comment (or tag) and hence remain displayed in draft state forever (or until the original creator re-opens them and adds a comment/tag).

Technically, Recogito deletes annotation if the user de-selects it, and it has no annotation bodies. 

One situation that would still lead to zombie annotations was the following:
- the user creates a new annotation
- without adding a comment or tag, the user closes the browser tab

This PR adds a listener to the page that will delete an empty, open annotation, if the user leaves the page.

Cf. issue https://github.com/performant-software/cove-recogito/issues/21

### DIFF
--- a/src/components/AnnotationDesktop/UndoStack/UndoStack.tsx
+++ b/src/components/AnnotationDesktop/UndoStack/UndoStack.tsx
@@ -1,10 +1,5 @@
-/**
- * Experimental code to a) prevent creation of empty annotations
- * and b) think about how we can implement undo functionality in 
- * the future.
- */
 import { useEffect, useRef } from 'react';
-import { Annotation, useAnnotationStore, useAnnotator, useSelection } from '@annotorious/react';
+import { Annotation, Annotator, useAnnotationStore, useAnnotator, useSelection } from '@annotorious/react';
 
 interface UndoStackProps {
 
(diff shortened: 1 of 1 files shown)


lwjameson (approved):
Nice.

SteveMarvin:
My logins don't seem to work for this preview.

lwjameson:
I see all of your various usernames there.  I assume you are getting 401's?

rsimon:
I have added an additional measure, attempting to prevent zombie annotations. (In the [@recogito/annotorious-supabase package](https://github.com/recogito/annotorious-supabase/commit/bb6470d2198075adebbfc972a4c535b2da04e397). It's now in the `develop` branch, i.e. not on the Netlify staging site yet!)

- Instead of using the Supabase JavaScript SDK, the call to archive annotations is now made through the REST interface directly.
- I'm setting the `keepalive` flag in the request, which instructs the browser to keep the request alive even after the page closes.

I'm not sure if this can address __all__ cases. (Behavior might differ between closing tab and actually closing the browser.) But from the client, I think this is pretty much as far as I can go. (FWIW: `navigator.sendBeacon` is said to be more robust than `fetch` + `keepAlive`. But the latter option does not seem to support adding custom HTTP headers, which we need for Supabase auth.

PR #95, opened 2023-11-23
Date Formatting

rsimon:
## In this PR

Changes the timestamp from [react-timeago](https://www.npmjs.com/package/react-timeago) to a GDocs-style absolute timestamp, using [react-moment](https://www.npmjs.com/package/react-moment). Supports EN and DE locales.

See: https://github.com/performant-software/cove-recogito/issues/20

Bildschirmfoto 2023-11-23 um 09 05 18
Bildschirmfoto 2023-11-23 um 09 05 01


absempere:
Do we still need to make a decision about this, or is it sorted?

> Hi @rsimon @lwjameson @absempere -- my view is that this should be configurable, per-organization at least. From a UX point of view, I agree with Rainer that the more "plain language" dates are more appropriate for most users. One project's preference for a more "precise" formatting notwithstanding. Let's discuss before merging!

rsimon:
I think we should ask Bonn about if they have a strong preference. Yes, it's a feature we could potentially make configurable. But it also feels like a fairly small detail. I'd be reluctant to make too many (small) things configurable. Any feature that can be toggled also introduces extra complexity to the codebase. (Especially in cases where toggling happens at runtime rather than compile time!) 

I.e. in particular for small things, I think it's ok to simply take a decision rather than trying to satisfy everyone. Personally, I always feel that if there's a good model or best practice you can point to, it's probably safe to do. GDocs will likely have thought through this. Following their lead won't be a mistake.

But as I said: we should ask Bonn's opinion on this.

absempere:
This sounds good -- thank you!

absempere:
Did we ever ask Bonn about this?

rsimon:
This one's good to merge now, too, right?

lwjameson:
Yes

PR #99, opened 2023-11-28
Redirect to pasted url when signed out after loging in

lwjameson:
### This PR does:

When a user pastes a recogito url and is not logged in, the url is stored and then after successful login is redirected to the pasted location.


rsimon:
Excellent addition - I agree this is something that was missing!

Just a question: I haven't used Nanostores in Astro before. I'm surprised that it works, because I thought Nanostores were supposed to be used only client-side. (To share state between Astro islands, but not across different requests.)

You are setting the URL on the Astro server, and I'm worried that this is actually a global state var. In fact, I tried the following:
- Open Chrome and Firefox (not logged in to Recogito in either)
- Paste the URL to a project by test user A into Chrome - this redirects me to the login page
- Now __without logging in to Chrome__, I go to FF `/sign-in` directly.
- Log in with test user B
- Test user B is redirected to the redirect URL from user A, and gets a 404 after login

rsimon:
P.S.: I noticed you created the helper component `/pages/redirect.astro`. Since this is a component, to be used inside another page, it should live either in `src/components`, or more likely in `src/apps/auth-login` somewhere. (Because that's the only page where it's used.)

The reason is: if you put it in `src/pages`, Astro will automatically create a route `/redirect` you can go to with the browser. With the current global store (and because the store isn't cleared after the redirect), going to `view-source:http://localhost:3000/redirect` now essentially provides a lookup for the last redirect in the system.

lwjameson:
One thing I was thinking about with the SSO problem was potentially using local storage to cache the redirect URL while the user is redirected to the IdP.

rsimon:
Yes, localStorage sounds good! And potentially more robust than anything based on a query param or HTTP header. (The various redirects that happen in the SSO flow might remove them, who knows.)

lwjameson:
@rsimon I have refactored this to use a combination of query parameters and local storage. There may be a couple of edge cases with the local storage part. For instance if you paste a url and then do not login and close your tab, the next time you login you will be redirected to the previously pasted url.  I do not think that is a huge issue.  If you can think of a better way let me know.

src/apps/auth-login/Login.tsx
```diff
@@ -34,21 +34,42 @@ export const Login = (props: {
 
   const [currentMethod, setCurrentMethod] = useState<LoginMethod | undefined>();
 
+  const url = new URLSearchParams(window.location.search);
+  let redirectUrl = url.get('redirect-to');
+  if (redirectUrl) {
+    localStorage.setItem('redirect-to', redirectUrl);
+  } else {
+    redirectUrl = localStorage.getItem('redirect-to');
+    if (redirectUrl && redirectUrl.length === 0) {
```
rsimon, line 43: I don't think this is strictly necessary. A string of length 0 is already 'falsy'.

src/apps/auth-login/Login.tsx
```diff
@@ -34,21 +34,42 @@ export const Login = (props: {
...
   useEffect(() => {
     supabase.auth.onAuthStateChange((event, session) => {
       if (event === 'SIGNED_OUT') {
         clearCookies();
       } else if (event === 'SIGNED_IN' || event === 'TOKEN_REFRESHED') {
         setCookies(session);
-        window.location.href = `/${props.i18n.lang}/projects`;
+        if (redirectUrl) {
+          window.location.href = redirectUrl;
+          localStorage.setItem('redirect-to', '');
```
rsimon, line 56: To clear, you can also use localStorage.removeItem('redirect-to').

rsimon (approved):
Looks good! Approved - I suggested two minor polishing things only.

P.S.: it's a bit of a pain that the redirect code needs to be repeated in each page (so so much else...) Astro didn't have the concept of a 'middleware' in the version we are using. I think that might have changed meanwhile. Hopefully there will be a chance at some point to migrate to the latest Astro version, clean things up and remove these redundancies.

rsimon:
P.S.: the edge case you're describing is a good one... (as with all edge cases where people close browsers in mid-workflow). Hm, one option, perhaps, is to add a handler on `onbeforeunload`, and clear the key/value pair in localStorage in the handler. (This should also get triggered when the user navigates away from the page - which means it could replace your other explicit `localStorage.removeItem()` statements.)

Not sure if there are other edge cases. (Can't think of any, off the top of my head.) But if there are, the only way is probably to give the redirect URL a max. lifetime: store the URL + store the creation timestamp. And then, when retrieving, check the age of the redirect URL.

PR #100, opened 2023-11-29
Branded Headers and Footers

rsimon:
## In this PR

This PR applies the branded Header/Footer components to all pages. 

__Caveat:__ the current CSS strictly requires 50px headers/footers. (We cannot remove them, without making additional CSS tweaks). I tried getting this to work with flexbox, but it seems that the custom `<astro-island>` components break the flexbox flow. I could not get the contents of the Astro island to stretch to the full height of the flex container.

__Therefore, if we want to support configuration with/without headers and footers in the future, we would need to move Header and Footer components out of the Astro layout, and into the React apps.__

__Another caveat:__ getting PDF to work was particularly tricky. I was not able to get the `fit page` option to work correctly. This option used to scale the PDF page to the height of the screen, resulting in very small text, thus probably making it the least useful option, anyway. I disabled the option, rather than spending too much time on debugging the issue.


lwjameson:
Hi @rsimon. I utilized the /src/dynamic-elements folder because  each tenant may have a different set of elements.  Instead of having to copy those elements into the layout folder, the dynamic-elements folder could be replaced by whatever was specified in the config.json during the CI/CD routines in the recogito-cloud repo. I suppose it can work this way as well, but it may be cleaner having the replaceable dynamic elements be in their own folder.  I doubt that this will be the last dynamic element replacement we will need eventually.

rsimon:
Ah - didn't know that. Yes, makes sense! Feel free to revert this change. (Should we call the folder "branding", perhaps?)

PR #101, opened 2023-11-30
Removed dependence on group names in the client

lwjameson:
# This PR adds

The recogito client was depending on specific names of groups to add users and such in various flows.  This PR removes this dependence.

All organization_groups, project_groups, and layer_groups now have is_admin and is_default columns that are set in the config.json for default groups.

So searches for "Org Admins" now uses a search for is_admin in the group.  "Project Admin" and "Project Student", and "Layer Student" is handled similarly.

Note that the Supabase support for this is only deployed on Bonn Staging for now.


rsimon (approved):
Looks good!

PR #102, opened 2023-12-01
Annotation Filtering

rsimon:
## In this PR

This PR adds the ability to filter annotations in the text and image annotation views. Filter categories are the same as for color coding:
- by assignment
- by annotation creator
- by privacy status (private vs. public annotation)
- by tag

Furthermore, this PR:
- Adds a little facelift to the design of the layers panel (which holds both the color-coding and the filter setting.
- Includes some refactoring to lift the color/filter state up, so that it remains persistent. (Previously, re-opening the layers panel would reset the color state back to default.)

Bildschirmfoto 2023-12-01 um 12 49 42


lwjameson (approved):
Looks good!

PR #104, opened 2023-12-12
Private documents

lwjameson:
## This PR

Implements private documents.  It requires [This PR](https://github.com/recogito/recogito-server/pull/7) to function so that may need to be approved and deployed first.

* All documents default to being Private
* In the document library you can set documents to Public. Doing so gives you a warning and a chance to decline beforehand.
  * The message is currently static in the I18n files but will be driven by the config file in Recogito Cloud deploys
* Private documents are only seen by the creating user in the Document Library


rsimon (approved):
Looks good!

The only thing I was wondering: do we need the new `ProjectDocument` type? I realize it's the type for the join table. But I was trying to keep the 'top level' types in `src/Types.ts` somewhat de-coupled from DB-specifics, and use them to model the 'domain'. (Especially in terms of providing compound types like `ExtendedProjectData`etc.)

The `ProjectDocument` type is only used as a utility inside the `backend` package, so my gut feeling is that it should live in there, but not at the level of the domain model. I.e. we could move it to the `backend/Types.ts` file, which already exists.

PR #105, opened 2023-12-18
Loading indicator overlay

rsimon:
## In this PR

This PR adds a simple, full-screen, semi-transparent loading overlay to the annotation views. The overlay displays a wait spinner, and makes the annotation UI underneath inert. It is removed as soon as the relevant resources for the annotation interface have loaded.

- For plaintext and TEI, loading is complete when the text/TEI content and annotations are fetched from the backend.
- For PDF, loading is complete when the PDF has been rendered and the annotations are fetched.
- For images, loading is complete as soon as the annotations are fetched.

### DIFF
--- a/package.json
+++ b/package.json
@@ -29,7 +29,7 @@
     "@radix-ui/react-toast": "^1.1.3",
     "@react-spring/web": "^9.7.3",
     "@recogito/annotorious-supabase": "^3.0.0-rc.3",
-    "@recogito/react-pdf-annotator": "^0.2.0",
+    "@recogito/react-pdf-annotator": "^0.3.0",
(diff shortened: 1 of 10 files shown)


package.json
```diff
@@ -29,7 +29,7 @@
     "@radix-ui/react-toast": "^1.1.3",
     "@react-spring/web": "^9.7.3",
     "@recogito/annotorious-supabase": "^3.0.0-rc.3",
-    "@recogito/react-pdf-annotator": "^0.2.0",
+    "@recogito/react-pdf-annotator": "^0.3.0",
```
rsimon, line 32: Added an `onRendered` callback to the `<PDFAnnotator>` component.

lwjameson (approved):
Looks good!

PR #108, opened 2023-12-20
Profile Nag

lwjameson:
## In this PR

This PR prompts the user to fill out their profile if they hit the Projects Dashboard and it has not been completed.  Requires first name, last name.


rsimon:
Made a few tweaks:
- Dialog now re-uses styles from the main stylesheet.
- Nag dialog button is semantically a link, so I changed from `<button>` to `<a>` which also simplifies the code a bit.
- Initial `open` state now determined directly without `useEffect`.

rsimon (approved):

PR #111, opened 2023-12-28
Support for collections

lwjameson:
## In this PR

The PR adds collections. Collections are a curated set of documents owned at the org layer.  Eventually we can add UI to allow Org Admins to create and curate collections, but initially these will be managed by server extensions.  The additions to the UI are in the Document Library:

image

Each collection adds a tab to the library. 

Collection documents can be updated, with previous versions retained.  Adding the document to a project always adds the most recent version. Previous versions can be selected by using the dropdown action menu:

image

The platform will allow multiple version of the same document in a project.

Another small update is that the document library now lists those documents currently in the project but disables the ability to uncheck (remove) them from a project.  This still must be done on the project dashboard.


rsimon:
It looks like I can't currently test this locally, because the Supabase cloud instance (against which I'm testing locally) doesn't have the `documents.collection_id` column yet. But overall, this looks good. Just two questions/remarks:

- There is some overlap between the tooltip styles and the popover-based tooltip that I'd like to resolve. (I.e. I'd like to align the styles, and move them into the global styles. A tooltip component would be a useful addition in other places as well.)
- Does the inclusion of the `collection_id` column mean that __every__ document needs to be part of a collection? Would this complicate the use case for Bonn? If collections are supposed to be an optional feature, would it make sense to keep collections out of the document schema, and instead have a separate association table instead?

rsimon:
I went ahead and made those CSS tweaks. (I.e. moved the tooltip styles into the global stylesheet. I also fixed some copy-and-paste issues: the Radix sample code includes vars that don't actually exist in our stylesheet.)

Let's chat about the way collections work in the next meeting. But for now, let's merge this PR & deploy it on the cloud instance for testing 👍

src/Types.ts
```diff
@@ -127,13 +127,20 @@ export interface Document {
 
   is_private: boolean;
 
+  collection_id: number;
```
rsimon, line 130: Does this mean that every document has to be part of a collection?

src/themes/default/tooltip/hover.css
```diff
@@ -0,0 +1,78 @@
+.tooltip-content {
```
rsimon, line 1: Note: I move those into the global stylesheet for re-use. Plus fixed:
- references to non-existing color variables used in the Radix demo
- missing animation keyframe settings

rsimon (approved):

src/Types.ts
lwjameson: No. collection_id is null for non-collection documents (any uploaded, added by users).

rsimon: Ah got it, thanks. 👍  In that case the type should reflect that `collection_id?: number`. I'll add that change later!

PR #112, opened 2024-01-03
Bugfix: Invalid 'IIIF_IMAGE' metadata value

rsimon:
## In this PR

This PR fixes the problem of file uploads getting an invalid `{ "protocol": "IIIF_IMAGE" }` value written into the `documents.meta_data` field. File uploads are now setting `meta_data` to undefined instead. (__Note:__ there's a not-null constraint on the field in Supabase, which will insert an empty `{}` object as a default.)

In addition, this PR adds a few missing German i18n labels.


lwjameson (approved):
Looks good

PR #115, opened 2024-01-09
Annotation sidebar filtering

rsimon:
# In this PR

This PR changes the behavior of the annotation list in the sidebar, so that it applies the same filtering as the document annotation view.

To make the user aware of the fact that a filter is being applied, the PR also adds a small notification bubble to the 'filter/layer' icon when a filter is active.

(In addition, the PR also fixes a small behavior bug, which caused the filter selections dropdown to get displayed with the "no filter active" option when opening the filter tab, even if a filter was currently active.)

Bildschirmfoto 2024-01-09 um 14 16 04

Bildschirmfoto 2024-01-09 um 14 15 21


lwjameson:
Hi Rainer. Though not an issue with this PR, I noticed that if you switch back and forth between the Annotation and Filter view on the sidebar that you lose the color-by setting item.  They are still color coded but the dropdown says "No color coding"

image

rsimon:
Argh - I fixed the exact same problem for the filter dropdown. And then failed to realize that it exists also for the color dropdown. Will send a separate PR for this today.

PR #116, opened 2024-01-10
Fix: color-coding dropdown no longer loses state when changing sideba…

rsimon:
## In this PR

This PR addresses [the issue reported here](https://github.com/recogito/recogito-client/pull/115#issuecomment-1883658700), which is that the color-coding dropdown lost state when switching back and forth between different sidebar tabs.


lwjameson (approved):
Looks good!

PR #117, opened 2024-01-10
Fix for who can edit/alter comments

lwjameson:
## In this PR

This is trying to tighten up some policy checks. It still does not work correctly but once we implement the add_document_to_project RPC this should start allowing Project Admins who are not Org Admins or Org Professors add documents to projects.


rsimon:
Ah great - so this will effectively "overlay" permissions, right? I.e. roll the actual permissions into one response, regardless of whether they originate from Org or Project-level policies?

lwjameson:
Yes. This should fix some issues with the UI not showing all available options.

PR #118, opened 2024-01-10
Add the ability to have dynamic CSS on collection documents

lwjameson:
## In this PR

Adds dynamic css capability to collections

### DIFF
--- a/src/Types.ts
+++ b/src/Types.ts
@@ -127,7 +127,7 @@ export interface Document {
 
   is_private: boolean;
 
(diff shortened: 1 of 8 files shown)


rsimon (approved):
I wonder if there isn't also a way to inject the CSS directly into the page template, at the Astro level. This would keep a bit of extra code out of the `TextAnnotationDesktop` component. (Which is actually getting a bit big for my taste, anyway.)

But this approach should definitely work - approved.

PR #120, opened 2024-01-16
Add warning when using toolbar

lwjameson:
## In this PR

This PR adds a warning when using the toolbar to delete another users annotation.


rsimon (approved):
Agreed, this works for now. It's not 100% ideal though, IMO that Toolbar now pulls a component that's embedded in the `PublicCommentActions`. Probably worth refactoring at some point. (Also, I'm wondering about whether we could replace the `<AdminOverrideAlert>` with the generic `<ConfirmedAction>` - which I think I created later...

PR #122, opened 2024-01-18
Addressing sorting crashes

lwjameson:
## In this PR

* Fix for this issue: https://github.com/performant-software/cove-recogito/issues/45

### DIFF
--- a/src/components/DocumentLibrary/DocumentLibrary.tsx
+++ b/src/components/DocumentLibrary/DocumentLibrary.tsx
@@ -553,8 +553,6 @@ export const DocumentLibrary = (props: DocumentLibraryProps) => {
           ),
         REVISION: (array) =>
           array.sort((a, b) => a.revision_count - b.revision_count),
-        LATEST: (array) =>
-          array.sort((a, b) => a.is_private.localeCompare(b.is_private)),
       },
     }
   );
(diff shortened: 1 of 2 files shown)


rsimon (approved):
I couldn't actually test this at the moment, because I don't have an instance with collections set up. (Although I guess it would be possible to point my local client to the COVE instance?)

Looks good overall. But see my comment below.

src/components/DocumentLibrary/DocumentLibrary.tsx
```diff
@@ -553,9 +552,7 @@ export const DocumentLibrary = (props: DocumentLibraryProps) => {
             (a.content_type || '').localeCompare(b.content_type || '')
           ),
         REVISION: (array) =>
-          array.sort((a, b) =>
-            a.item.revision_count.localeCompare(b.item.revisions?.length)
-          ),
+          array.sort((a, b) => a.revision_count - b.revision_count),
         LATEST: (array) =>
           array.sort((a, b) => a.is_private.localeCompare(b.is_private)),
```
rsimon, line 557: Given that `is_private` is a boolean, wouldn't this line still fail?

lwjameson: Oh.  That is no longer used. I went ahead and removed it.

PR #126, opened 2024-01-21
Replaced Logo Placeholders

rsimon:
## In this PR

This PR makes a small change that's been bothering me for a while. Since we don't have a Recogito logo, I had originally put two dummy placeholders into the UI:
- A generic "Kanban" icon in the project dashboard
- The Google Podcasts logo (!) in the project overview sidebar :-) 

Obviously, we can't go live with the Google Podcasts logo. I therefore replaced both of these placeholders with yet another generic - but at least unproblematic - placeholder. We should probably think about a logo at some point. But for now, this PR adds an MIT-licensed generic "annotations" icon. 

Bildschirmfoto 2024-01-21 um 10 07 11

Bildschirmfoto 2024-01-21 um 10 07 27


src/layouts/project/ProjectSidebar.tsx
```diff
@@ -73,30 +73,30 @@ export const ProjectSidebar = (props: ProjectSidebarProps) => {
               <NavItem 
                 tabIndex={-1}
                 className="no-hover"
-                icon={GooglePodcastsLogo}
+                icon={<RecogitoCircleLogo className="recogito-logo-circle" />}
                 label='Recogito'
```
lwjameson, line 77: Can we change this to 'Projects'? I think it is more meaningful for navigation purposes.

lwjameson (approved):
Looks good. I have one suggestion in my comments.

src/layouts/project/ProjectSidebar.tsx
rsimon: Agreed. I wonder if we should get rid of the logo altogether in this case, and just use a "back" caret, labeled "Projects"?

PR #128, opened 2024-01-31
Open Join and Open Edit

lwjameson:
## In this PR

This PR reworks a number of aspects of creating new projects and adds new features for projects in general.

### Project Creation Dialog

* Creating a new project via the Create Project button now presents a user with a dialog to specify the parameters of the project.

image

  * This dialog requires both name and description to be filled in before Save is activated
  * The info circles will put up an tooltip explaining the feature 
  * Once created you are immediately taken to the Project's home page.

* Open Join - Open join projects allow all registered users to join your project without an invitation
* Open Edit - Allows all members of the project to add annotations to the default context

* The project creation process has been moved to a server side RPC.

* Added column is_project_default to contexts table, eliminating dependence on checking for a null name or DEFAULT_CONTEXT tag to identify.

### Project Settings updated

* You can now adjust most aspects of the project from the Settings page.

image

### Misc fixes

* Added an "Add an annotation..." placeholder to the Rich Text Editor.

image

* Updated the Profile page to be more descriptive.

image


lwjameson:
@rsimon I have deployed the server requirements to recogito-staging.covecollective.org server. I can also deploy it to the Bonn staging server if that would be easier for you to review.

src/apps/dashboard-projects/ProjectsHome.tsx
```diff
@@ -55,18 +55,31 @@ export const ProjectsHome = (props: ProjectsHomeProps) => {
...
-    !me.last_name.length);
+  const [showProfileNag, setShowProfileNag] = useState(false);
 
   useEffect(() => {
     getMyProfile(supabase).then(({ error }) => {
       if (error) window.location.href = `/${props.i18n.lang}/sign-in`;
     });
   }, []);
 
+  useEffect(() => {
```
rsimon, line 66: `me` is coming in as a prop on mount, and wouldn't change I believe. Is there a need for the extra effect hook?

src/Types.ts
```diff
@@ -172,6 +176,8 @@ export interface Context {
   description?: string;
 
   project_id: string;
+
+  is_project_default: boolean;
```
rsimon, line 180: +1. We used to do this via tagging. Does this mean the tag-based approach for identifying the default context is no longer needed?

src/apps/project-home/ProjectHome.tsx
```diff
@@ -71,7 +69,7 @@ export const ProjectHome = (props: ProjectHomeProps) => {
 
   const { addDocumentIds } = useDocumentList(
     project.id,
-    defaultContext!.id,
+    defaultContext?.id,
```
rsimon, line 72: Just for my understanding: can the defaultContext ever be undefined? (Or would this point to an integrity problem in the DB?)

src/apps/project-settings/ProjectSettings.css
```diff
@@ -34,4 +54,70 @@
   display: flex;
   gap: 0.2em;
   align-items: center;
-} 
\ No newline at end of file
+}
```
rsimon, line 57: We should definitely move those general styles (for the switch) into the global stylesheet to avoid repetition further down the line.

src/apps/project-settings/ProjectSettings.tsx
```diff
@@ -1,70 +1,116 @@
 import { useEffect, useState } from 'react';
-import { clearProjectTagVocabulary, getProjectTagVocabulary, setProjectTagVocabulary } from '@backend/helpers';
+import {
```
rsimon, line 2: We should definitely also refactor this at some point. Project access settings and the tag vocabulary are separate things, and we should pack them into separate components. Perhaps wait for Chelseas review first, but we should IMO put it on the list.

src/backend/helpers/documentHelpers.ts
```diff
@@ -187,6 +195,7 @@ export const listDocumentsInProject = (
         return { error, data: [] };
       } else {
         // Simplify layers from list of contexts to single (default) context
+        console.log(data);
```
rsimon, line 198: Leftover debug log :-)

src/components/CreateProjectDialog/CreateProjectDialog.tsx
```diff
@@ -0,0 +1,125 @@
+import * as Dialog from '@radix-ui/react-dialog';
```
rsimon, line 1: If this is only used on the dashboard, I wonder if we should just move it into `apps/dashboard-projects/CreateProjectDialog`?

src/components/InfoTooltip/InfoTooltip.css
```diff
@@ -0,0 +1,95 @@
+.info-tooltip-content {
```
rsimon, line 1: There are already some general CSS tooltip styles in `themes/default/tooltip`. I.e. much of this might not be needed. Also to point out that the demo styles from the Radix docs are referencing color vars we don't actually have!

I believe it should actually be fine to simply remove this stylesheet altogether. As long `tooltip`, `tooltip-arrow` and `tooltip-content` classes to your tooltip, it should pick up the base styles we already have.

src/components/InfoTooltip/InfoTooltip.tsx
```diff
@@ -0,0 +1,25 @@
...
+}
+export const InfoTooltip = (props: InfoTooltipProps) => {
+  return (
+    <Tooltip.Provider>
+      <Tooltip.Root>
+        <Tooltip.Trigger asChild>
+          <Info className='info-tooltip-icon' />
+        </Tooltip.Trigger>
+        <Tooltip.Portal>
+          <Tooltip.Content className='info-tooltip-content' sideOffset={5}>
```
rsimon, line 17: Re above: I think if you replace `info-tooltip-content` with `tooltip-content` and `info-tooltip-arrow` with `tooltip-arrow`, our base styles should kick in.

rsimon (approved):
Looks good! Left a few comments & questions, but really mostly for tweaks and possible future refactoring.

src/apps/dashboard-projects/ProjectsHome.tsx
lwjameson: Probably my over zealous/SPA bent where I never trust a prop on mount.

src/Types.ts
lwjameson: I believe so yes. When looking at the code, you almost always find the default context by checking if its name is null. I actually did not see anywhere where the code check the tag.  If you know of a spot let me know.

src/components/InfoTooltip/InfoTooltip.tsx
lwjameson: I will make that change.

src/apps/project-home/ProjectHome.tsx
lwjameson: Yes. If the user is not a member of a layer in the default context, then it is null as it does not pass RLS.  It hasn't been a problem I guess since that user should. not see any documents anyway, but if use a student user in Production you will see an error thrown every time they go into the project home.

src/components/CreateProjectDialog/CreateProjectDialog.tsx
lwjameson: Yeah.  I have not really always made the Only used in this app, general purpose component.  I am not sure we have been totally consistent on this.

src/components/InfoTooltip/InfoTooltip.css
lwjameson: I have tended to add the custom CSS.  I do believe I have changed the vars to match ours when we actually used the class, but I see I left some unused classes from the demo :-/

lwjameson:
@rsimon I will probably need you to update the german translations before I merge.

PR #129, opened 2024-02-01
Annotation view fixes

rsimon:
## In this PR 

This PR adds a few fixes to the annotation view.

- Re-adds the PDF scale controls to the new menu bar
- Fixes PDF layout that got broken during the layout redesign
- Annotation list sidebar: when the list overflows vertically, the sidebar no longer gets unwanted horizontal scrollbars.
- Text annotation view: 
   - Mousewheel scrolling now works on the entire page (it was broken when the mouse hovered over the page background, rather than the text page itself)
   - The scrollbar for the text panel is now actually grab-able (d'oh - it wasn't actually usable before due to a buggy `pointer-events: none` setting)
- Cleanup: some CSS styles + components from the old menu layout that are no longer used were removed


lwjameson (approved):
Looks good!

PR #130, opened 2024-02-01
Responsive Reply Form

rsimon:
## In this PR

This PR attempts to add a mobile/responsive reply form to the annotation interface.  

- The reply form now has a "mobile fallback" mode, which spawns a React portal with a full-screen textarea.
- Whether the fallback is used or not is tested when the RichTextEditor gets focus.
- At the same time, a resize listener is attached to to the window. For each resize event, it is tested whether the fallback textarea should be used instead of the Quill RTE.
- The resize listener is removed again in the following situations:
   - Quill fires the onBlur event (it seems, however, this doesn't happen reliably)
   - The ReplyForm gets unmounted
- The test evaluates window height. Currently, a fixed threshold of 512px is used. If the window height falls below, the mobile fallback is activated.

The expected sequence of events is:
- User clicks into Rich Text Editor.
- RTE gets focus, virtual keyboard pops in, triggering resize events.
- Fallback kicks in as soon as the screen is smaller than the threshold.

__Chrome iOS Issue:__ when closing the fallback textarea, the whole annotation UI gets pushed to the left side for some reason I do not yet understand yet. I have vague hopes that this might be the Netlify deploy preview wrapper intervening. (But probably not...) The same issue does not seem to happen on Chrome Android. (Tested with my phone.)

__For later improvement:__ I have used a plaintext textarea as the fallback for now. We can change this to a Quill RTE later, once we know that the approach works.


lwjameson (commented):
I am seeing some pretty odd behavior on an iPad.  We should probably have Steve take a run at this first on a selection of devices before merging this.

PR #131, opened 2024-02-12
User management

lwjameson:
## In this PR

This PR adds user management for Org Admins.

image

image

An Org Admin can now Manage registered users, changing their Org level role, or deleting them. For the time being I have not added a 'Create User' function as that will take some discussion. As most of our current users enter via some form of SSO, creating email based users from this interface is problematic and may cause confusion.

### DIFF
--- a/src/apps/user-management/DeleteUser/DeleteUser.tsx
+++ b/src/apps/user-management/DeleteUser/DeleteUser.tsx
@@ -1,0 +1,115 @@
+import { useState } from 'react';
+import * as Dialog from '@radix-ui/react-dialog';
+import { Trash, X } from '@phosphor-icons/react';
+import { Button } from '@components/Button';
+import type { PostgrestError } from '@supabase/supabase-js';
+import { removeUserFromProject } from '@backend/crud';
+import { supabase } from '@backend/supabaseBrowserClient';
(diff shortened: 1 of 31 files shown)


src/pages/[lang]/users/index.astro
```diff
@@ -0,0 +1,64 @@
+---
```
Comment (rsimon) [inline src/pages/[lang]/users/index.astro:1]: There are still a few unused imports and a debug log statement.

src/apps/user-management/DeleteUser/DeleteUser.tsx
```diff
@@ -0,0 +1,115 @@
...
+  const name = nickname
+    ? nickname
+    : first_name || last_name
+    ? [first_name, last_name].join(' ')
+    : undefined;
+
+  const onDelete = () => {
+    setBusy(true);
+
+    removeUserFromProject(supabase, user.id, user.org_group_id).then(
```
rsimon, line 42: This doesn't really do anything on the project level, correct? Worth renaming this helper method to `removeUserFromGroup` instead?

rsimon (approved):
Looks good! I had some minor CSS issues with not being able to scroll my users table fully to the bottom. (The end position of the scroll was the bottom of the browser window, and the footer obscured the last table row.) But we can sort this out later.

PR #135, opened 2024-02-16
TEI Behaviors

rsimon:
## In this PR

This PR adds the following custom CETEIcean behaviors:

- `rend` and `rendition` attributes on `<figure>` elements are copied to the generated HTML ``.
- `width` and `height` attributes on `<graphic>` elements are copied to the generated HTML ``.
- `<desc>` elements are hidden.
- The `innerHTML` of the `<desc>` is copied into the `alt` attribute of the generated HTML ``.


lwjameson (approved):
Looks good!

PR #136, opened 2024-02-16
More reasonable project options for Project Card

lwjameson:
## In this PR

* Changes the requirements for deleting a project to Org Admin or Project Owner
* Uses the new leave_project_rpc to actually leave project (old leave project failed because user did not have DELETE privs on the group_users table.


rsimon (approved):
Looks good! 

P.S.: I like that we're moving more things to RPCs. We should probably write up some documentation about this at some point. Is there a way, perhaps, to pull something out of the auto-generated documentation that the Supabase dashboard generates?

lwjameson:
I will look at what is possible in terms of documentation.

PR #137, opened 2024-02-19
Conditional TEI stylesheet import

rsimon:
## In this PR

This PR adds a bit of extra logic to the TEI annotation page: if the TEI document is associated with a custom CSS stylesheet, the TEI annotation page now no longer includes the default TEI CSS styles.

Before, the TEI annotation view either had __both__ stylesheets (TEI default styles + custom stylesheet), or only the default stylesheet. Now it's __either__ default styles, __or__ custom stylesheet, but not both.


lwjameson (approved):
Looks good.

PR #138, opened 2024-02-19
Text Annotator Library Update

rsimon:
## In this PR

This PR upgrades the following libraries.

- @recogito/react-text-annotator
- @recogito/react-pdf-annotator
- @recogito/annotorious-supabase

__Proceed with caution!__ The latest text annotator includes a breaking change in the annotation data model. A fix to patch the difference on the fly was included into annotorious-supabase. But keep your eyes open for irregularities.


lwjameson (approved):
Well lets give it a go!

PR #141, opened 2024-02-20
Enhancements: CSS Highlight Rendering, Message Throttling

rsimon:
## In this PR

This PR upgrades to the latest versions of `@recogito/react-text-annotator`, `@recogito/react-pdf-annotator` and `@recogito/annotorious-supabase` and enables the following new features:

- __Experimental CSS Custom Highlight rendering in text and PDF annotation views.__ This eliminates the annotation drift problem on newer browsers. (On older browsers that don't support the CSS Custom Highlight API, rendering will fall back to Canvas rendering, i.e. no change to previous behavior.)
- Note: CSS Custom Highlight rendering is __not__ enabled for TEI yet, since there seem to be issues with TEI custom elements on some browsers. (Safari on all platforms, Chrome on iOS.)
- __Realtime message throttling.__ Messages sent across the Supabase realtime channel are now throttled. A maximum of 10 messages/second is sent per client. During the 100ms buffering period, annotation change events are aggregated and buffered. This means __fast user actions are no longer silently discarded by Supabase!__ This should make our realtime sync a lot more robust.


lwjameson (approved):
Not much for me to review, but the preview looks good.

PR #143, opened 2024-02-26
Redirect to join project

lwjameson:
## In this PR

The client allows users to send direct links to users to access a specific annotation context.  Currently when a user is not a member of the project, they receive a 404 Not Found. This update checks wether the project is Open Join/Open Edit.  If so the user is presented a UI which allow them to join the project:

image

If they choose Join, they will be joined to the project and redirected to the annotation they were trying to access.

If the project is not Open Join/Open Edit the behavior is the same as before.

### DIFF
--- a/src/pages/[lang]/annotate/[context]/[document]/index.astro
+++ b/src/pages/[lang]/annotate/[context]/[document]/index.astro
@@ -31,7 +31,6 @@ if (document.error || !document.data) {
   // Lets check if this is an open join project
   const resp = await isOpenJoinEditFromContext(supabase, contextId);
   if (resp) {
-    console.log('is open!');
     return Astro.redirect(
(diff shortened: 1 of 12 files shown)


lwjameson:
@rsimon This is testable against COVE Staging.

src/pages/[lang]/annotate/[context]/[document]/index.astro
```diff
@@ -27,6 +28,16 @@ if (!contextId || !documentId) {
 // Will return empty if this user has no access priviliges
 const document = await getDocumentInContext(supabase, documentId, contextId);
 if (document.error || !document.data) {
+  // Lets check if this is an open join project
+  const resp = await isOpenJoinEditFromContext(supabase, contextId);
+  if (resp) {
+    console.log('is open!');
```
Comment (rsimon) [inline src/pages/[lang]/annotate/[context]/[document]/index.astro:34]: Debug log :-)

rsimon (approved):
Looks good!

PR #144, opened 2024-02-29
IIIF Presentation Manifest Support (Beta)

rsimon:
## In this PR

This PR adds (beta!) support for importing IIIF presentation manifests.

#### Annotation Interface
- Basic __multi-image annotation support__. When in multi-image mode, each annotation now includes the URI of the image as a `source` parameter in the selector. The client still downloads all annotations for the whole document (!) - but filters based on the `source`.  (See caveats and future enhancements below!) 
- A left drawer component with a __thumbnail strip__. 
   - The thumbnail strip is virtualized (using [react-window](https://github.com/bvaughn/react-window)). 
   - The current image is marked with a blue outline. Clicking a thumbnail opens the image in the annotation view.
   - The thumbnail labels are pulled from the presentation manifest.

Bildschirmfoto 2024-02-29 um 11 38 28

#### Project Dashboard
- __Modified 'Import IIIF' dialog__. The IIIF import dialog now performs, first, a basic validation of the URL format. If the URL is technically valid, it is resolved and content downloaded as JSON. If successful, the JSON is passed through IIIF validation (using [@allmaps/iiif-parser](https://github.com/allmaps/allmaps/tree/main/packages/iiif-parser)), which: 
   - identifies the type of manifest (image vs. presentation)
   - extracts the document label (in case of a presentation manifest)
- __Modified drag & drop handler__. The same validation process is also applied when dragging-and-dropping a URL into the project dashboard.
- When saving the document record to Supabase, the __manifest label (if any) is used as document name__, and the __protocol metadata field__ is set to `IIIF_IMAGE` or `IIIF_PRESENTATION`, according to the IIIF validation result.
 
Bildschirmfoto 2024-02-29 um 11 45 33

### Caveats
- __Testing!__ So far I only tested with a few resources from the Bodleian Library and the BNF. In theory both v2 & v3 (of the IIIF Image as well as Presentation API) should be supported. So far have I only found v2 resources to test, however.
- __Multi-image annotation support is still a bit tacked on__. When selecting a new image from the thumbnail sidebar, what happens is this: the `Annotator` (= the Annotorious component in charge of managing the annotation lifecycle) is destroyed and re-initialized. This has the following negative consequences:
   - All annotations (for the whole document, not just the current images!) are downloaded again from Supabase
   - The presence manager is re-created, which a) causes a flicker in the presence avatar list, and b) will assign new colors to each present user

To address the multi-image problems, multi-image support would have to be built more deeply into the Annotorious architecture. IMO we should look into this after the basic todos are ticked off the list, and we know if (and how much) budget is left.

### Possible Future Enhancements

It would be nice to see in the sidebar:
- which images have annotations (and how many)
- potentially, which users are currently looking at which image. (Similar to how presence works in Google Slides, for example.)

Both features, however, would be easier if we first implement "native" multi-image support in Annotorious, rather than patch around it. I.e. Definitely a heavier lift.


lwjameson (approved):
Looks good.

PR #145, opened 2024-02-29
Save Error Indicator

rsimon:
## In this PR

This PR adds an error alert to the annotation UI, which pops up on Supabase connection or annotation save errors. Because connection errors are fatal, the alert also blocks the annotation UI and prompts the user to reload the page.

Bildschirmfoto 2024-02-29 um 16 46 37

P.S.: this also addresses issue: https://github.com/performant-software/cove-recogito/issues/61#issuecomment-1971167443. (If the user is logged out, this would cause a connection error. Refreshing the page would redirect to the login page.)


lwjameson (approved):
Looks good!

PR #150, opened 2024-03-18
Google analytics

lwjameson:
## In this PR

- Add header tags for Google Analytics
- Fix node js Astro config


rsimon (commented):
Hm, do we really want to hard-wire the Google Analytics script into every page? IMO we should at least make the import conditional. We could make it so that if no Google API key is set in the `.env` file, the script wouldn't be imported at all.

Also, a word of warning: AFAIK, Recogito will strictly need a cookie popup if we use Google Analytics. (I believe the rules say that you don't strictly need a cookie popup if the only thing you are using them for is your own session tracking. But as soon as any analytics or 3rd parties are involved, EU cookie popup is needed.)

lwjameson:
I had considered a conditional, but much like the hardwired UserSnap, nothing works unless we define a tag id for Google right? Otherwise nothing happens.  I assume we will only turn this on for places that want it.

rsimon:
It wouldn't work if you don't set a `PUBLIC_GOOGLE_ANALYTICS_TAG_ID` key in the .env file, true. But it will now fire a request to the Google Tag manager on every page load, which might potentially make some sys admins uneasy. Wouldn't something like the below work in the Astro template?

```html
{Boolean() && (
    <script
      async
      src=`https://www.googletagmanager.com/gtag/js?id=${import.meta.env.PUBLIC_GOOGLE_ANALYTICS_TAG_ID}`
    ></script>
    <script>
      // @ts-ignore
      window.dataLayer = window.dataLayer || [];

      function gtag() {
        // @ts-ignore
        dataLayer.push(arguments);
      }
      // @ts-ignore
      gtag('js', new Date());
      // @ts-ignore
      gtag('config', `${import.meta.env.PUBLIC_GOOGLE_ANALYTICS_TAG_ID}`);
    </script>
)}
```

lwjameson:
Your right.  We should probably do the same thing with UserSnap.

rsimon (approved):

PR #158, opened 2024-03-25
Improvements to splash backgrounds

lwjameson:
## In this PR

- Improved implementation of the Login background to use higher resolutions if available


rsimon:
Looks good!

PR #161, opened 2024-03-27
Client Support for Supabase/Cantaloupe-based IIIF

rsimon:
## In this PR

This PR adds support for an alternative IIIF setup, available when self-hosting Recogito.
- Instead of uploading images to IIIF Cloud, it is now possible to upload to Supabase storage directly.
- It is then assumed that the self-hosted setup includes a Cantaloupe image server that has Supabase storage configured as its data source.

To configure the desired setup, use the following two new environment variables:

```env
# Either IIIF_CLOUD or SUPABASE_CANTALOUPE
PUBLIC_IIIF_CONFIGURATION="IIIF_CLOUD"

# The (absolute or relative) base path of the Cantaloupe server.
# IIIF requests will go to ${PUBLIC_IIIF_CANTALOUPE_PATH}/${document_uuid}
PUBLIC_IIIF_CANTALOUPE_PATH="/iiif" 

```


rsimon:
Just to add:
- Jann has confirmed this is working with the local Cantaloupe setup
- I have made the IIIF Cloud-based configuration the default. If you don't set any of the new `.env` vars, things will keep working as is, and none of the changes in this PR should affect the cloud instances.

lwjameson (approved):
Looks good!

PR #168, opened 2024-04-09
Read only layer support changes and New UX

lwjameson:
## In this PR

This is an very large PR with a lot of changes both in appearance and the interaction between client and the backend. 

- Should match current level of functionality.
- Requires the backend support for the changes in the policy system
- Implements the UX changes for project-home
- Implements different views for project admins and students
- Many backend manipulations are replaced with RPC's 

This PR need thorough testing of all current functionality.


src/pages/[lang]/annotate/[context]/[document]/index.astro
```diff
@@ -66,13 +65,17 @@ const channelId = data.digest('hex');
 
 let viewer;
 
+console.log('Document: ', document.data);
```
Comment (rsimon) [inline src/pages/[lang]/annotate/[context]/[document]/index.astro:68]: Leftover debug logging

src/pages/[lang]/projects/[project]/assignments/index.astro
```diff
@@ -47,3 +47,4 @@ const i18n = getTranslations(Astro.request, 'project-assignments');
     documents={documents.data}
   />
 </ProjectLayout>
+@apps/project-home/AssignmentsView
```
Comment (rsimon) [inline src/pages/[lang]/projects/[project]/assignments/index.astro:50]: Leftover copy-and-paste artefact? :-)

rsimon (approved):
It's good to see that much as simplified with the introduction of the RPCs!

Let's go a head and merge this, and then clean up on details later.

rsimon:
P.S.: now that the project sidebar is gone, we need a place to put the plugin configuration. Wonder whether you (or Chelsea) have any thoughts on that. Perhaps somewhere underneath "Project Settings" would make sense. IMO we should generally allow for a second hierarchy level in project settings, covering different topics (access/visibility, tag vocabs, plugins, etc.)

rsimon:
I merged the other commits and then fixed conflicts with this branch. The checks are still failing, but as far as I can see, the problem is just the remaining Astro 'Assignments' page that's no longer pointing anywhere:

```
src/pages/[lang]/projects/[project]/assignments/index.astro
```

It includes the following import, that no longer exists:

```
import { ProjectAssignments } from '@apps/project-assignments';
```

I'm hesitant to make any changes in there myself. But if I understand correctly, that whole page is no longer needed, right?

src/backend/Types.ts
```diff
@@ -20,3 +20,9 @@ export interface ProjectDocument {
 
   document_id: string;
 }
+
+export type userRole = {
```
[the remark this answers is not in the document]
lwjameson: This is received from the server. I have moved this to src/Types

PR #174, opened 2024-05-06
Annotation Card UX & read-only layer support in annotation views

rsimon:
## In this PR

This PR implements the  new UX for the annotation cards in popup and list mode, as well as the new card design for 'notes'.

In addition, there is now support for read-only annotations in the text- and image annotation views.
- Card style changes according to editable/read-only status
- Text highlight style is either underline (for read-only) or color highlight
- Image annotation style is either unfilled & black stroke (for read-only) or semi-transparent color

As part of the new annotation card UX, I also rebuilt the rich text editor component. It's now based on Quill directly (rather than `react-quill`). This eliminates a bunch of lifecycle & focus issues, as well as deprecation warnings caused by `react-quill` apparently using an outdated Quill version underneath the hood. There's also a new custom toolbar, with our own icons & style.


lwjameson (approved):
A lot here, but seems to run well.  Like getting rid of React Quill...It was a pain.

Approved!

PR #176, opened 2024-05-14
Revised UX for Annotation Views 

rsimon:
## In this PR

This PR implements the majority of the new UX for the text/PDF and image annotation views. Note that some smaller issues are still missing. But I think it makes sense to merge what we have now. Todos that remain open (and will follow by the end of the week):

- Color legend popover.
- ~~Auto-suggest dropdown for tags.~~ __Done__
- The annotation Delete button that used to be next to the floating panel is still missing. (It is possible to delete from the annotation card, though.)
- ~~The popup is currently disabled when the right sidebar is open. However, it should only be disabled if the user is looking at the "Annotations" tab, not the notes tab.~~ __Done__
- Style changes are needed for the "My Account" icon (needs to move to the header bar) and the presence avatar stack.
- ~~(Almost) none of the labels are localized yet!~~ __Done__
- There are no undo/redo buttons yet. (Although functionality-wise, this is now available in Annotorious. I'd need to double-check for the Text Annotator.)
- Plugin extension points are no longer available in the new UX. (I can still cover this within the OU contract.)

P.S.: I have also updated various library dependencies in the `package.json` file. I only upgraded minor versions, as to not break anything. But eventually, there are some major version upgrades (most importantly Astro!) we should look into.


lwjameson (approved):
Looks good!

PR #179, opened 2024-05-16
Annotation Topbar

lwjameson:
## In this PR

- Added the Top Bar to the annotation view

### DIFF
--- a/src/apps/annotation-image/ImageAnnotation.tsx
+++ b/src/apps/annotation-image/ImageAnnotation.tsx
@@ -1,26 +1,26 @@
 import { Annotorious } from '@annotorious/react';
-import {
-  PluginProvider,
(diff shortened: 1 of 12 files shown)


rsimon (approved):
Looks good!

I made some changes to the CSS though. Using fixed pixel heights, hard-wired margins, in combination with `calc`, `position: fixed`, etc. is the path to the dark side ;-) As soon as the first future change arrives, things start to become untraceable. And it's not needed in the annotation interface, because the normal `flex` layout flow can handle the annotation view layout just fine.

P.S.: I also tweaked the padding of the header to align with the toolbar. All my changes are overrides to the annotation view specifically, though! I.e. nothing should affect the other pages that use the `TopBar`.

PR #180, opened 2024-05-17
More Annotation View Enhancements

rsimon:
## In this PR

Adds more enhancements to the annotation view.

- Notes now have read-only layers support.
- Adds the layer icon to read only Annotation and Note cards.
- Adds an 'unread notes' notification bubble to the 'Notes' tab item.
- When creating a new Note, the UI now uses the new UX design (https://github.com/performant-software/cove-recogito/issues/80)
- Fixed a bug that cause the new UI to crash when a note was received async, while the Notes panel was open
- Made minor tweaks to the Notes card styles
- Fixed a bug in the text annotation view that caused the "presence cursor" (= the label on the annotation when a collaborator had an annotation selected) to disappear when scrolling


lwjameson (approved):
Looks good!

PR #182, opened 2024-05-22
More UX bug fixes

lwjameson:
## In this PR

Addressed the following issues:

- https://github.com/performant-software/vico/issues/259
- https://github.com/performant-software/vico/issues/265
- https://github.com/performant-software/vico/issues/258
- https://github.com/performant-software/vico/issues/257

Also addressed a tricky issue where being removed as a user of an Open Join project meant that you would not see that project again. The fix implemented a RPC that removed a user from a project and any contexts.

### DIFF
--- a/src/Types.ts
+++ b/src/Types.ts
@@ -1,5 +1,4 @@
 export interface UserProfile {
-
   id: string;
 
   nickname?: string;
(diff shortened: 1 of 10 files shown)


src/Types.ts
```diff
@@ -160,65 +139,51 @@ export interface Document {
...
-
   context: DocumentContext;
-  
-  layers: DocumentLayer[];
 
+  layers: DocumentLayer[];
 }
 
 export interface DocumentContext {
+  id: string | undefined;
```
rsimon, line 165: If they are optional, you can also use `id?: string`. I believe those two notations are identical, as far as the effect on auto-completion goes. Just mentioning it because in this particular case we have both notations in the same interface.

rsimon (approved):

rsimon:
Looks good!

PR #183, opened 2024-05-22
UX bug fixes

lwjameson:
## In this PR

Addresses the following issues:

- https://github.com/performant-software/vico/issues/266
- https://github.com/performant-software/vico/issues/261


src/components/AnnotationDesktop/FilterPanel/AnnotationLayers/AnnotationLayers.tsx
rsimon, line 1: I'm not really sure about this change. Is this something that Chelsea recommended? Personally, I'd still prefer the standard filter dialog, even if there's only one layer, I think.

I guess we could discuss whether layers are enough of a "power feature" that many people wouldn't use or be aware of in the beginning. In which case a layers dialog that only ever has one layer might look confusing. But even then, I'd probably stick with "Annotation Layers" as a heading, and then maybe just have "Show Annotations" as a label. Or just "Annotations" as a heading, and "Show annotations" as a label.

Either way, the current wording of "Show Annotation" -> "Show" feels kind of odd & redundant to me, as does the inconsistency with the multilayer project. (Different heading, different icon.) Maybe something to discuss quickly on Friday?

rsimon (changes requested):
Looks good overall, but may be we can quickly chat with Chelsea tomorrow about the change to the layers menu in the annotation?

rsimon (approved):

PR #185, opened 2024-05-24
More Annotation View UX Fixes

rsimon:
## In this PR

- Fixes https://github.com/performant-software/vico/issues/275
- Implements alternative design for view/hide annotations when there's only a single layer (cf. https://github.com/performant-software/vico/issues/266))


lwjameson (approved):
Looks good.

PR #189, opened 2024-06-03
Fixes for usage of Cantaloupe IIIF

lwjameson:
## In This PR

- Fixes for using a cantaloupe configuration.


rsimon (approved):
Looks good.

I looked through the commits. I don't think that's actually a regression - but seems to have never worked... Jann may have only tested up to the point of uploading the image perhaps. 

Thanks!!

PR #191, opened 2024-06-05
Fix for Cantaloupe loaded images

lwjameson:
## In this PR

Images loaded from a Cantaloupe based IIIF server behind Supabase requires a token, the current approach to loading IIIF images with an image tag fails.  This PR instead gets a signed URL from Supabase storage to display the image.

### DIFF
--- a/src/apps/annotation-image/IIIF/useIIIF.ts
+++ b/src/apps/annotation-image/IIIF/useIIIF.ts
@@ -1,11 +1,12 @@
 import { useEffect, useState } from 'react';
 import { Utils, Manifest, Resource, Sequence } from 'manifesto.js';
-import type { DocumentInTaggedContext } from 'src/Types';
+import type { DocumentWithContext } from 'src/Types';
 import { supabase } from '@backend/supabaseBrowserClient';
 
 type ManifestType = 'PRESENTATION' | 'IMAGE';
 
-const CANTALOUPE_PATH: string | undefined = import.meta.env.PUBLIC_IIIF_CANTALOUPE_PATH;
+const CANTALOUPE_PATH: string | undefined = import.meta.env
+  .PUBLIC_IIIF_CANTALOUPE_PATH;
 
 // Performs a simple sanity check
 const isSupported = (manifest: Manifest) => {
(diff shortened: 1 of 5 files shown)


rsimon (approved):
Damn... yes, of course. Hadn't thought about that. It's not possible to just use the IIIF URL in an `` tag. Getting the signed image URL is the correct way to go.

The only issue I'm seeing is that this will now download the full-size image, right? The image could massive, so we should resize on the server. I found this:

https://supabase.com/docs/guides/storage/serving/image-transformations#signing-urls-with-transformation-options

This also says that image resizing is available on "Pro plan and above" only. So I wonder what that means for the self-hosted setup? If that's not an option, we could still go through Cantaloupe. In this case, we'd need to fetch the image through an AJAX request (so we can add auth headers), and then create the image tag programmatically. Let me know if you want to try that, and I can give it a shot. (I've implemented something like this previously.)

jsmueller7:
@rsimon Probably my mail didn't go through, so I'll repeat it here:

In order to perform these kind of image transformations you must also run imgproxy. I haven't deployed that, but it is possible to include into a self hosting environment. Still it seems like shooting sparrows with cannons since cantaloupe should be able to do image transformations as well. There should be an easy way to retreive thumbs...

rsimon:
Ok, sounds sensible. Yes, I agree it's not worth deploying an extra component for the image resizing, given that we already have Cantaloupe. Building a slightly smarter Thumbnail component that fetches the image async, with auth header attached, would be a cleaner solution. @lwjameson let me know if you want me to take a look.

lwjameson:
Sure @rsimon. Give it a go.  I wanted to make it fairly simple, but if you have a good way of downloading and displaying an in-memory image. I was just short on time and wanted to fix the issue that @jsmueller7 was having. In my current self-hosting setup I do have imgproxy deployed so we could take that route as well but do not want to create complications for Jann's setup.

rsimon:
Will do - can look at it later today.

jsmueller7:
Reteiving a reseized image from Cantaloupe isn't actually too hard. A Cantaloupe URL works like this:
`/{identifier}/{region}/{size}/{rotation}/{quality}.{format}`. In order to get an thumbnail image the url could be `https://<my-recogito-instance>/iiif/2/<image-UUID>/full/!100,100/0/default.jpg`. This way you'll get a thumb.

lwjameson:
Understand the IIIF syntax, the complication being that you cannot use this in an image tag as we need credentials to access the Cantaloupe service.  All doable with AJAX calls, but certainly more complicated.

lwjameson:
@rsimon Are you sure we will need that?  To my knowledge we will not be supporting IIIF Presentation Manifest upload via Recogito so I would assume that all Presentation manifests from public IIIF servers will work fine as is.

rsimon:
Good point... yes you are right! I think at the moment the "Pages" tab is always shown, even for a single image. But if so, this should be disabled, anyway. I'll check and fix if it's the case.

PR #194, opened 2024-06-11
Implemented expanded project description

lwjameson:
## In this PR

- Implemented an Show More / Show Less buttons for long project descriptions.
- Addresses issue: https://github.com/performant-software/vico/issues/277

### DIFF
--- a/src/apps/project-home/ProjectHeader/ProjectHeader.css
+++ b/src/apps/project-home/ProjectHeader/ProjectHeader.css
@@ -21,6 +21,22 @@
   font-weight: 400;
   font-size: 14px;
   line-height: 22.4px;
+  display: block;
+  display: -webkit-box;
+  line-clamp: 3;
+  line-height: 160%;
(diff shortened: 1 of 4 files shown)


rsimon (approved):
Let's merge this for now. But TBH the page layout isn't ideal. 

The immediate problem: the button overlaps with the text. It can work sometimes (if the text doesn't fill the last line). But looked like this in my case:

Bildschirmfoto 2024-06-12 um 06 56 50

I tried to tweak it, but than ran into problems. (I wrapped the description into a `<p>`, so I could apply line clamp and add some bottom padding to make room for the button. Then realized that the header section started overlapping the "Documents" label, because it has fixed position and a hard-wired margin. Then didn't dare to change the margin because it's switched dynamically depending on admin/non-admin.) TBH I think we really need to refactor the layout at some point, to scale properly with the content. Otherwise all those hard-wired pixel heights and fixed positions are going to bite us on different devices in the future.

PR #195, opened 2024-06-13
Invite Users

lwjameson:
## In this PR

- Adds the ability for Org Admins to invite email/password users from the User Management screen.
   - Fulfills: https://github.com/performant-software/cove-recogito/issues/81

- Adds two new client environment variables:
  - PUBLIC_ENABLE_USER_INVITE
    - If set to TRUE, it enables this feature (on client and server)
    - FALSE or undefined disables the feature
  - SUPABASE_SERVICE_KEY
    - If PUBLIC_ENABLE_USER_INVITE is TRUE, this is required by the invite-user API endpoint to invoke the invite user function

  - These have been enabled on the Staging site.

### DIFF
--- a/src/apps/user-management/InviteUserDialog/InviteUserDialog.css
+++ b/src/apps/user-management/InviteUserDialog/InviteUserDialog.css
@@ -1,0 +1,134 @@
+.invite-user-dialog-overlay {
+  background-color: var(--black-a9);
+  position: fixed;
+  inset: 0;
+  animation: overlayShow 150ms cubic-bezier(0.16, 1, 0.3, 1);
+}
+
+.invite-user-dialog-content {
+  background-color: white;
+  border-radius: 6px;
+  box-shadow: hsl(206 22% 7% / 35%) 0px 10px 38px -10px,
+    hsl(206 22% 7% / 20%) 0px 10px 20px -15px;
+  position: fixed;
+  top: 50%;
+  left: 50%;
+  transform: translate(-50%, -50%);
+  width: 90vw;
+  max-width: 450px;
+  max-height: 85vh;
+  padding: 25px;
+  animation: contentShow 150ms cubic-bezier(0.16, 1, 0.3, 1);
+}
+
+.invite-user-dialog-content:focus {
+  outline: none;
+}
+
+.invite-user-fieldset {
+  display: flex;
+  gap: 20px;
+  align-items: center;
+  margin-bottom: 15px;
+}
+
+.invite-user-dialog-title {
+  margin: 0;
+  font-weight: 500;
+  color: var(--mauve-12);
+  font-size: 17px;
+}
+
+.invite-user-dialog-description {
+  margin: 10px 0 20px;
+  color: var(--mauve-11);
+  font-size: 15px;
+  line-height: 1.5;
+}
+
+.invite-user-button {
+  display: inline-flex;
+  align-items: center;
+  justify-content: center;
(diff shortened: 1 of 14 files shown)


src/apps/user-management/InviteUserDialog/InviteUserDialog.css
```diff
@@ -0,0 +1,134 @@
+.invite-user-dialog-overlay {
```
rsimon, line 1: There's actually no need for `.invite-user-dialog-overlay`. If you just set the class to `.dialog-overlay`, it will pick up the default styles from the main stylesheet.

src/apps/user-management/InviteUserDialog/InviteUserDialog.css
```diff
@@ -0,0 +1,134 @@
...
+  background-color: var(--black-a9);
+  position: fixed;
+  inset: 0;
+  animation: overlayShow 150ms cubic-bezier(0.16, 1, 0.3, 1);
+}
+
+.invite-user-dialog-content {
```
rsimon, line 8: Same here. Simply set the class to `dialog-content`, and the main stylesheet will kick in.

src/apps/user-management/InviteUserDialog/InviteUserDialog.css
```diff
@@ -0,0 +1,134 @@
...
+  left: 50%;
+  transform: translate(-50%, -50%);
+  width: 90vw;
+  max-width: 450px;
+  max-height: 85vh;
+  padding: 25px;
+  animation: contentShow 150ms cubic-bezier(0.16, 1, 0.3, 1);
+}
+
+.invite-user-dialog-content:focus {
```
rsimon, line 24: Also defined in the main stylesheet, if you use `dialog-content` as a class.

src/apps/user-management/InviteUserDialog/InviteUserDialog.css
```diff
@@ -0,0 +1,134 @@
...
+}
+
+.invite-user-fieldset {
+  display: flex;
+  gap: 20px;
+  align-items: center;
+  margin-bottom: 15px;
+}
+
+.invite-user-dialog-title {
```
rsimon, line 35: Also possible to just use `dialog-title` and leave the rest up to the main stylesheet.

src/apps/user-management/InviteUserDialog/InviteUserDialog.css
```diff
@@ -0,0 +1,134 @@
...
+  margin: 0;
+  font-weight: 500;
+  color: var(--mauve-12);
+  font-size: 17px;
+}
+
+.invite-user-dialog-description {
```
rsimon, line 42: Likewise: there's a `dialog-description` class.

src/pages/api/invite-user.ts
```diff
@@ -0,0 +1,78 @@
...
+    );
+  }
+
+  if (!orgAdminResp.data) {
+    return new Response(JSON.stringify({ error: 'Unauthorized' }), {
+      status: 401,
+    });
+  }
+
+  // Org Admin, Go ahead and use the Service Key to invite
```
rsimon, line 61: I guess there's no way around using the service key here, right?

rsimon (approved):

src/pages/api/invite-user.ts
lwjameson: Afraid not.  That API requires it.

PR #197, opened 2024-06-19
Astro: Migration from v2.10 to v4.10 (among other things)

rsimon:
## In this PR

This PR upgrades Astro from v2.10 (which we started with 1.5 years ago) to the latest version, 4.10. In addition, this PR upgrades pretty much all other dependencies to their latest versions (including the TypeScript version), and fixes various breaking changes resulting from the upgrades.

Additionally, this PR:
- adds DE labels that were missing (mostly in the Document Library, User Admin and the updated Assignments Wizard)
- fixes some minor typing issues left over from the introduction of the read-only layers feature
- cleans up the code base a bit, removing unused code and imports

Additonal notes:
- `react-moment` is no longer supported, and was beyond repair with the latest TypeScript and Vite version. Date formatting was replaced with `date-fns`.


lwjameson (approved):
Nice. Looks good.

PR #200, opened 2024-06-21
Regression Issue

lwjameson:
## In this PR

- Fixes a regression issue with invite user


rsimon (approved):
Ah - thanks for spotting. I missed those.

PR #201, opened 2024-06-21
Fix for avatar overflow

lwjameson:
## In this PR

Currently project cards do not limit the number of avatars shown:

image

This PR does this:

image

### DIFF
--- a/src/components/ProjectCard/ProjectCard.tsx
+++ b/src/components/ProjectCard/ProjectCard.tsx
@@ -84,6 +84,12 @@ export const ProjectCard = (props: ProjectCardProps) => {
     });
   };
 
+  const count = users.length;
+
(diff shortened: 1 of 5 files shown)


src/components/ProjectCard/ProjectCard.tsx
```diff
@@ -84,6 +84,12 @@ export const ProjectCard = (props: ProjectCardProps) => {
     });
   };
 
+  const count = users.length;
+
+  const userList = count < 11 ? users : users.slice(0, 10);
```
rsimon, line 89: Theoretically, you could just use `userList = users.slice(0, 10);` here. (It won't complain if the list < 11). Not that it makes much of a difference though ;-)

rsimon (approved):
Looks good - I added the German 'more' label.

PR #203, opened 2024-06-23
Netlify Runtime Issue: Updated Build Config

rsimon:
## In this PR

This PR resolves runtime errors on Netlify. I frankly don't know why our own library (`@recogito/annotorious-supabase`) needed to be added to `noExternal` after the Astro 4.1 upgrade. But without it, Netlify crashed at runtime with the error message `crypto is not defined`.


lwjameson (approved):
Lets do it.

PR #206, opened 2024-06-25
Bugfix: Text Annotation Filtering

rsimon:
## In this PR

Fixes a bug in the text annotation interface. When hiding annotations with filters, the hidden annotations were still selectable, and would cause the mouse cursor to turn to "hover" state when moving over them. Note that the bug was fixed in the `text-annotator` base library here:

- https://github.com/recogito/text-annotator-js/commit/4315f612ffb8d40e05e45fa24cb2f5a875c8933d
- https://github.com/recogito/text-annotator-js/commit/c18a261bca5562194b61e59ecd051243a1cd1a65

P.S.: the mouse cursor issue still persists with the latest published `text-annotator` version. I noticed it after I released `3.0.0-rc.33`. I fixed the bug, but didn't want to publish another release just for this.


lwjameson (approved):
Not much to review here. Approved!

PR #207, opened 2024-06-25
Join Requests

lwjameson:
## In this PR

This PR adds support for Join Requests. Join requests allow an admin to create a link that they can send or post for users to request membership.  The general flow is as follows:

- Admin copies join request link:
image

- If that link is clicked or pasted into the browser you are navigated to /projects/<project_id>/request:
image

- Clicking the `Request` button creates the request and gives feedback of success:
image
  - Clicking OK sends you to the Projects page

- When the Admin logs in, the Team button shows a pip if there are active Join Requests:
image

- The team page shows any active join requests.  You can also choose to show previous request that they have ignored:
image

- Clicking accept adds the user to the team with the default role:
image

The Staging server has been updated with the need support for testing.


src/apps/project-request/RequestResultMessage.tsx
```diff
@@ -0,0 +1,55 @@
...
+  state: 'success' | 'failure';
+}
+
+export const RequestResultMessage = (props: RequestResultMessageProps) => {
+  const { open } = props;
+  const { t } = props.i18n;
+  return (
+    <AlertDialog.Root open={open}>
+      <AlertDialog.Portal>
+        <AlertDialog.Overlay className='request-result-overlay' />
```
rsimon, line 21: Is there a reason you needed to define custom styles for this? There are global dialog styles in `src/themes/default`. Normally, you just would need to use the appropriate classes (`dialog-overlay`, `dialog-content`, `dialog-title`,  `dialog-description`) and the stylesheet should handle the reset. Or am I missing something?

src/pages/[lang]/projects/[project]/request.astro
```diff
@@ -0,0 +1,39 @@
...
+    client:only='react'
+    i18n={i18n}
+    projectId={projectId as string}
+    user={profile.data}
+    isAlreadyMember={project.data}
+  />
+</BaseLayout>
+
+<style>
+  h1 {
```
Comment (rsimon) [inline src/pages/[lang]/projects/[project]/request.astro:36]: I don't think this will do anything, since the Astro styles are scoped. If you want the h1 to apply on the page globally, you'd need to use either

```
<style is:global>
  h1 { ... }
</style>
```

or 

```
<style>
  :global(h1) { ... }
</style>
```

or (better IMO) move it into the stylesheet of the React component.

rsimon (commented):
Looks good. I left a few comments on minor things. Just one additional note:

I initially tested this on an Open Join project. This, then lead to a page that told me I'm already a member. This is technically correct, I guess. But seemed confusing. Would it  be better to simply hide the join link on open projects?

src/apps/project-collaboration/ProjectCollaboration.css
```diff
@@ -31,3 +31,43 @@
   line-height: 170%;
   padding: 10px 0;
 }
+
+.project-collaboration-request-header {
+  width: 50%;
```
rsimon, line 36: The `width: 50%` keeps collapsing things for me. Especially the text in the "copy link" button gets wrapped and then exceeds the button height.

rsimon (approved):

jsmueller7:
> Looks good. I left a few comments on minor things. Just one additional note:
> 
> I initially tested this on an Open Join project. This, then lead to a page that told me I'm already a member. This is technically correct, I guess. But seemed confusing. Would it be better to simply hide the join link on open projects?

Maybe. The most important thing for us is that you can place a link in our LMS; in principle, you can already do this with an Open Join project. The function must work for of a lecturer who does not understand exactly how the project configurations work in detail and who may also not understand that you could simply take the URL from the address field if you take an "Open Join" project.

It might also be interesting to see what happens if you change the access control (from Open Join to Join-Reqest). Are there different links for this, or does the behavior change when you click on the link?

jsmueller7:
> I added german language labels and fixed a small bug in `CreateProjectDialog.tsx` I noticed. (Clicking the label for the lower buttons would change the state in the upper buttons, because lower and upper buttons had the same ID.)

Great! There is this small thing that you can hardly call an issue but maybe fits in: If you close the create project dialouge without actually creating a project, the create project button keeps showing this "loading circle" until you refresh the dashboard.

src/apps/project-request/RequestResultMessage.tsx
lwjameson: Eh. Sorry. Just a copy/paste thing from elsewhere.  I will clean it up.

src/pages/[lang]/projects/[project]/request.astro
Comment (lwjameson) [reply src/pages/[lang]/projects/[project]/request.astro:36]: Agreed. Honestly I need to go back a clean up some things before merging. I will update the PR.

lwjameson:
I will also disable the whole Join Request section for Open Join projects.  It already works that if you navigate to an open join project (through a link) then it asks you if you want to join and if so, joins you and redirects to wherever the link led.

src/apps/project-request/RequestResultMessage.tsx
lwjameson: Actually AlertDialog and Dialog are different. This classes do not work for this.

src/apps/project-collaboration/ProjectCollaboration.css
lwjameson: Fixed!

PR #208, opened 2024-06-26
New Recogito Studio branding

lwjameson:
## In this PR

Switched out both the Favicon and added the Recogito Studio logo to TopBar. This PR is mostly to get some screen shots and I suspect that we will need to do further work before merging to develop.


rsimon (approved):
Yay! 🎉 Let's deploy it :-)

PR #210, opened 2024-06-28
Display Embedded TEI Standoff Annotations as Read-Only Layers

rsimon:
## In this PR

This PR adds support for displaying annotations embedded in TEI `<standOff>` elements as read-only layers. Each standoff element is presented as a separate layer.

Also see related additions to the `text-annotator` base library here:
- https://github.com/recogito/text-annotator-js/commit/024d9a311fde87b2c3b5f4409fe79d57cae942f7
- https://github.com/recogito/text-annotator-js/commit/6bd59bf5ab8570e0e61d172124bd1d2fc8caf557

Additionally, I made a few minor fixes & enhancements :
- TEI export used to fail when the document name contained unicode characters. When exporting, the filename is now sanitized and slugified.
- The `src/Types.ts` file still contained some inconsistencies for Layer types. I cleaned this up a little (hopefully not breaking anything...), introduced a new `EmbeddedLayer` type, and removed some left-over functions in `src/backend` that are no longer used.
- Fixed a minor display issue in the annotation list: layer tooltips were cut off on (semi-transparent) un-selected cards. I now disabled the tooltip on un-selected cards altogether, since this seemed more sensible.
- Updated all library dependencies to latest versions, where possible (i.e. not uppy or OpenSeadragon)

Note: due to the way the `text-annotator` works, sorting of embedded annotations is currently an unsolved issue. This will come in a future PR (and will require more changes to the text annotator library first).


lwjameson (approved):
Looks good!

PR #215, opened 2024-07-03
Minor Bugfixes

rsimon:
## In this PR

Fixes three small issues:
- Images embedded in Quill rich text are now scaled to a maximum size (100% width of the editor or max. 0.3 * viewport height, whichever limit is encountered first)
- Fixes the wrong label reported in https://github.com/performant-software/vico/issues/299
- Addresses the issue reported by the ATRIUM project, that IIIF image manifests with query params (`.../info.json?foo=bar`) were not opened in the annotation view.


lwjameson (approved):
Looks good.

PR #217, opened 2024-07-11
Migrated IIIF handling from manifesto.js to @allmaps/iiif-parser

rsimon:
## In this PR

This PR migrates all IIIF handling code from the (brittle...) `manifesto.js` to the `@allmaps/iiif-parser`.


lwjameson (approved):
Looks good.

PR #220, opened 2024-07-11
Document library loading changes

lwjameson:
## In this PR

- It was reported that newly added documents were not appearing in My Documents in the document library. It turns out that we are now over 5000 documents on Staging and the limit per query is 5000. This change loads documents in batches.


rsimon (approved):
Hm, with multiple 1.000 docs in the library already, I feel we should schedule a more fundamental redesign of this component at some point. We'll ultimately need lazy loading, and server-side support for different sorting, search, etc. to handle large libraries in a useful way.

For now, that looks good. But lets bring it up in one of the next planning meetings.

PR #221, opened 2024-07-12
Layer De-Duplication

rsimon:
When the annotation UI mounts, document layer provided through `props.document.layers` are now de-duplicated. This is a temporary fix until we resolve the root cause of our duplicate layers.


lwjameson (approved):
Looks good.

PR #226, opened 2024-07-22
Dashboard CSS/Behavior Fixes

rsimon:
## In this PR

This PR adds a number of style and behavior fixes to the main project dashboard.

- Fixed 'busy' button spin icon behavior on the initial, empty dashboard.
- Grid/Table view toggle button: the Radix Toggle Group component allows for an unselected state. (I.e. neither the Grid nor the Table button is selected.) Added a fix that now properly emulates radio button behavior, so that either Grid or Table is selected.
- Minor cosmetic tweaks to the table view and owner pill.
- Minor cosmetic tweaks to the user account menu's admin section.
- Cleanup: removed redundant button styles on the "public project" card and "join project" dialog, and used global theme styles.
- On Chrome, buttons with an icon only (such as 'X' close buttons) were only clickable if you hit the icon, but not the padding around it (even though the padding is inside the HTML `<button>` component). This is fixed now.


lwjameson (approved):
Looks good!

PR #228, opened 2024-07-24
IIIF View Improvements

rsimon:
## In this PR

This PR makes small enhancements to the image annotation view. See also: https://github.com/performant-software/vico/issues/229

* In case of a multi-page IIIF presentation manifest, the thumbnail sidebar is open initially.
* In case of a single image, there is no more "Filters / Pages" tab. Only the filter panel is shown. (Previously, the Pages tab was available, but always empty.)


lwjameson (approved):
Looks good.

PR #231, opened 2024-07-25
Account Settings Facelift

rsimon:
## In this PR

This PR provides a facelift to the Account Settings page.

- Added the branding TopBar component.
- Localized button labels.
- Removed the "Back" link at the top in favour of "Cancel" and "Save" buttons.
- Tweaked DE translation.
- Behavior: "Save" now saves & navigates back immediately. User no longer stays on the page after save.
- When going back (either by clicking "Cancel" or "Save") the browser now goes to the actual previous location, not to `/{lang}/projects`. (The hard-coded back link wasn't a problem previously. But after the UX changes, the user can now access the account settings page from anywhere across the app.)

There's also a bit of cleanup, removing unused code.

Bildschirmfoto 2024-07-25 um 09 42 13

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -108,19 +108,27 @@
       }
     },
     "node_modules/@annotorious/annotorious": {
-      "version": "3.0.0-rc.29",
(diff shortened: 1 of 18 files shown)


lwjameson (approved):
Fantastic! I had been thinking about doing this for awhile.

PR #232, opened 2024-07-25
Rudimentary Rollback After Failed File Upload

rsimon:
## In this PR

This PR adds a basic rollback mechanism that removes a document record, in case a subsequent file upload fails. It's not perfect though and could be improved, I think:

* In case of a failed upload, the `removeDocumentsFromProject` is called.
* This, in turn, uses the `archive_project_documents_rpc` RPC.
* Therefore, both the document record as well as other data (e.g. a row in `project_documents`, maybe more?) remain in the DB.


lwjameson (approved):
Looks fine. I will try and make sure that everything is cleaned up.

PR #233, opened 2024-07-26
'Fatal Error' Dialog for Annotation Pages

rsimon:
## In this PR

This PR adds a React error boundary around the text-/image-annotation interface, to display a (minimal, but at least reasonably useful) alert dialog if anything goes wrong when mounting the annotation UI.

Bildschirmfoto 2024-07-26 um 11 34 26


lwjameson (approved):
Good stuff. We may want something similar on other pages.

PR #234, opened 2024-07-30
Annotorious + Text Annotator: Updates to Latest Versions

rsimon:
## In this PR

There have been some breaking changes recently in Annotorious and the Text + PDF Annotator libraries. The main breaking change is a rewrite of the text annotation popup, which has been contributed by a 3rd party. Placement behavior seems better than before, but I had to make some bugfixes. 

This also changes the PDF annotator to use the `legacy` build, which should support (slightly) older browser as well.


lwjameson (approved):
Looks good.

PR #235, opened 2024-07-30
Fix labels and descriptions

lwjameson:
## In this PR

I missed changes to the Baselayer name and description described [here](https://github.com/performant-software/vico/issues/285#issuecomment-2226359681). Will need german translations


rsimon (approved):
Looks good! I added DE translations.

PR #239, opened 2024-08-01
Fix for needing to refresh to see layers from newly added assignments

lwjameson:
## In this PR
- Fix for https://github.com/performant-software/vico/issues/296
- Really bad bug in the assignment wizard

### DIFF
--- a/src/apps/project-home/AssignmentsView/Wizard/Progress/ProgressUpdating.tsx
+++ b/src/apps/project-home/AssignmentsView/Wizard/Progress/ProgressUpdating.tsx
@@ -104,7 +104,7 @@ export const ProgressUpdating = (props: ProgressUpdatingProps) => {
         // - check for added read only layers on new documents
         for (let i = 0; i < documentChanges.added.length; i++) {
           const doc = documentChanges.added[i];
-          const prevDoc = previous.documents.find((d) => (d.id = doc.id));
+          const prevDoc = previous.documents.find((d) => d.id === doc.id);
           const layerDiff = diff(prevDoc!.layers, doc.layers);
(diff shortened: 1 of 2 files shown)


src/apps/project-home/AssignmentsView/Wizard/Progress/ProgressUpdating.tsx
```diff
@@ -104,7 +104,7 @@ export const ProgressUpdating = (props: ProgressUpdatingProps) => {
         // - check for added read only layers on new documents
         for (let i = 0; i < documentChanges.added.length; i++) {
           const doc = documentChanges.added[i];
-          const prevDoc = previous.documents.find((d) => (d.id = doc.id));
+          const prevDoc = previous.documents.find((d) => d.id === doc.id);
```
rsimon, line 107: Such a classic :-D (Having fond memories of doing the same thing at various times ;-)

src/apps/project-home/ProjectHome.tsx
```diff
@@ -64,6 +69,12 @@ export const ProjectHome = (props: ProjectHomeProps) => {
     }
   }, [isAdmin]);
 
+  useEffect(() => {
```
rsimon, line 72: Since `ProjectHome` gets only rendered once, with the initial props, I guess you could skip the effect, and just initialize the state from the props?

I.e.

```ts
const [availableLayers, setAvailableLayers] = 
  useState(props.availableLayers);

// ...

/*
useEffect(() => {
  if (props.availableLayers) {
    setAvailableLayers(props.availableLayers);
  }
}, [props.availableLayers]);
*/
```

PR #245, opened 2024-08-20
CSV download for all project users

rsimon:
## In this PR

Addresses the requirement that any project member (admin or not) should be able to download annotations as CSV.

@lwjameson: I could use a sanity check on this. Before, I only allowed download for (project/org) admins. I have changed the permissions check now. Anyone who has `SELECT` privileges on the `project_documents` table can download. Is this a correct way to infer document/annotation read access?


lwjameson (approved):
Looks Good! Yep. project_documents works!

lwjameson (changes requested):
Woops. Actually let me check that more closely.

lwjameson (changes requested):
The problem with this is that currently the export CSV is only available on the Documents page, which non-admins would never see. This may require some design.

rsimon:
There is a menu dropdown option available on the document cards. This is also available in the Assignments view. But, true, the expectation would probably be that this should only download annotations for this particular assignment...

rsimon:
Ok, I looked into this again.
- Only admins can see the "Download All CSV" button (because it's on the document view)
- Non-admins, however, always got to see the "Download CSV" dropdown option that's in the document three-dot context menu.
- However, this option would always return an "unauthorized" error - this works now, with this PR

Here's what I would suggest:
1. We introduce an additional URL param to the export endpoint, so that the assignment is taken into account. Right now, even if you export from within the assignment, you'd get all annotations (from the entire project) visible to you, which is not what I'd expect as a user.
2. We _could_ do a redesign, so that the Assignments tab also gets a "Download All CSV" button, which would give you all annotations, in this assignment.

I haven't heard 2. raised as a requirement directly. But feel it might be included implicitly in what the client wants? (CCing @jsmueller7)

lwjameson:
I think 1. should be sufficient for now, although since this is off the document card it would seem odd to get all document annotations when intuitively you are really only looking for the annotations for that document. We could try:

- Add URL parameters for context_id and document_id.
- Add a button to the Assignment detail pane that lets you download all annotations for the assignment:

image

rsimon:
Yes, that's what I meant. The context menu option already takes the document into account. I.e. option 1 would be to also take into account the context ID, thus giving you the annotations for this document, in this assignment. Option 2, as you indicated in your screenshot, would be that additional button to get all annotations from all documents - in this context.

rsimon:
Phew... that ended up quite the yak shaving 😀 I realized we actually already had assignment-level downloads previously! But they got erased when we eliminated the `/projects/{id}/assignments/{id}` URL route with the new UX.

Everything is back as it should be now:
- Project-level CSV export from the Documents View works as before
- At the Assignment level, there's now a similar "Download All" button, as sketched by you above, @lwjameson. This will download all annotations, on all documents, from all layers in this assignment. The button is available to every project member, not just admins.
- Document card context menu:
   - In the Document View, the CSV/TEI download options will download all annotations in the project, on this document
   - In the Assignment View, the options will download the annotations on this document, from all layers in this assignment.*

* Note that we could limit to the active layer if we want. But for now, I have included all layers in the context.

lwjameson (approved):
Looks good!

PR #246, opened 2024-08-20
localStorage Persistence for Document Homepage State

rsimon:
## In this PR

This PR makes the following two state variables in the Project Home view persistent via localStorage:
- Current tab (either "Documents" or "Assignments")
- Currently selected Assignment

If the user reloads the page, or returns from the annotation view, they will end up at the same tab, and with the same Assignment selected as they last left the page.

__Note:__ it's not a perfect solution though. If you leave the Project Home page via the "Back to Projects" link, you will later return to the persisted state. From a user perspective, you'd probably expect to enter into the initial state ("Documents" tab). In the long run, we should probably persist the state through URL params instead. This way, we can:
- Maintain state when reloading the page,
- Have a clean state when linking from the dashboard,
- Link back to a specific state from the annotation view.

### DIFF
--- a/src/apps/project-home/AssignmentsView/AssignmentsView.tsx
+++ b/src/apps/project-home/AssignmentsView/AssignmentsView.tsx
@@ -1,4 +1,4 @@
-import { useEffect, useState } from 'react';
+import { useMemo, useState } from 'react';
 import { GraduationCap } from '@phosphor-icons/react';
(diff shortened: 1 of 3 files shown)


lwjameson (approved):
Looks good. I agree we may need to implement query params eventually since they may be useful for annotation linking as well.

PR #251, opened 2024-09-04
Add emails to invite list of users

lwjameson:
## In this PR

- Previously the "Add list of users' functionality did not send email. This PR updates so that it sends the same email as the 'Add user' button.


rsimon (approved):
Looks good!

PR #254, opened 2024-09-16
PDF Export

rsimon:
## In this PR

This PR adds export of PDF files with embedded Recogito annotations. PDFs can be exported at the Admin level ("Documents" tab), or the Assignments level ("Assignments" tab). In both cases, it's possible to choose between including or excluding private annotations. 

Unfortunately, the PDF standard only supports a very limited annotation model. Therefore, the following restrictions apply:

- There are no clickable links. Embedded [linked text](http://www.example.com) is serialized to plaintext: `linked text (https://www.example.com)`.
- Embedded Base64 images are dropped.
- For images or videos embedded via URL, only the URL is serialized.
- Tags are serialized as `[MyTag] [MyOtherTag]`.
- Replies to a comment are separated by a line, and the the creator name & timestamp is added before the reply text.

#### Screenshot: Exported PDF in Acrobat Reader

Bildschirmfoto 2024-09-16 um 12 43 15


lwjameson (approved):
Looks good!

PR #255, opened 2024-09-16
Color Legend Bugfix 

rsimon:
## In this PR

This PR fixes a bug in the annotation view. The color legend did not update live when annotations changed. This caused, for example, the following problem:

- User switches to color coding "by first tag".
- User creates a new annotation, with a tag that was not used before.
- The color legend won't show this tag (until the user changes the color coding to something else and then back).

The current fix could be improved for performance: when annotations change, __all__ color legends are re-computed. We could revisit this later and clean up, so that only the visible legend is re-computed. But for now, this works.


lwjameson (approved):
OK

PR #256, opened 2024-09-16
Refactoring/bugfix: unread notes

rsimon:
## In this PR

This PR fixes a small regression. In image annotation mode, the little notification bubble that shows the number of unread notes (see image below) was no longer shown. (It worked in text annotation mode, but did not use the shared component from the `AnnotationDesktop` components.)

This PR adds some refactoring, so that both image- and text-annotation mode are using the shared component again.

Bildschirmfoto 2024-09-16 um 13 44 04


lwjameson (approved):
Looks good.

PR #258, opened 2024-09-20
Fix for Bonn ENV var issue

lwjameson:
## In this PR

Bonn is having issues with Invite Users to Project.  They were able to get email working process.env vs import.meta.env for passing ENV vars to the API.

However process.env does *not* work when running locally. This PR uses either method to try and ingest the vars.

* Additionally a bug with inviting a single user was addressed.

### DIFF
--- a/src/pages/api/invite-user-to-project.ts
+++ b/src/pages/api/invite-user-to-project.ts
@@ -22,8 +22,6 @@ export const POST: APIRoute = async ({ request, cookies, url }) => {
   // Verify if the user is logged in
   const supabase = await createSupabaseServerClient(cookies);
 
-  console.log('Host: ', MAIL_HOST);
-
   const me = await getMyProfile(supabase);
 
   if (me.error || !me.data)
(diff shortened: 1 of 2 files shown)


src/pages/api/invite-user-to-project.ts
```diff
@@ -9,10 +9,21 @@ import {
...
+const MAIL_PASSWORD =
+  process.env.MAIL_PASSWORD || import.meta.env.MAIL_PASSWORD;
+const MAIL_FROM_ADDRESS =
+  process.env.MAIL_FROM_ADDRESS || import.meta.env.MAIL_FROM_ADDRESS;
+
 export const POST: APIRoute = async ({ request, cookies, url }) => {
   // Verify if the user is logged in
   const supabase = await createSupabaseServerClient(cookies);
 
+  console.log('Host: ', MAIL_HOST);
```
rsimon, line 25: Debug log :-)

rsimon:
Ok, I guess there's no reason we shouldn't just allow both versions, so this looks like a safe option.

> However process.env does not work when running locally. 

Do you mean it doesn't work when running with Vite in dev mode?

rsimon (approved):

lwjameson:
> Ok, I guess there's no reason we shouldn't just allow both versions, so this looks like a safe option.
> 
> > However process.env does not work when running locally.
> 
> Do you mean it doesn't work when running with Vite in dev mode?

Yes.

PR #259, opened 2024-09-23
Default Language Setting

rsimon:
## In this PR

This PR:
- makes the default language configurable via ENV var.
- applies the default language of message localization in the `invite-user-to-project` API endpoint.


lwjameson (approved):
Looks good.

PR #262, opened 2024-09-24
SAML redirect fix

lwjameson:
## In This PR

Fixes issue where SAML logins are not redirected to the original URL.


jsmueller7:
I've tried this out by integrating the `Login.tsx` file into my local setup. The effect is that I am redirected as I hit the SSO-Login-Button, but the SSO flow isn't started, so I am not directed to the SP page where I can supply my credentials.

PR #264, opened 2024-09-25
Locked Projects

lwjameson:
## In this PR

UI and backend work for Locked projects


src/apps/project-settings/ProjectSettings.css
```diff
@@ -28,6 +28,8 @@
   max-width: 65ch;
   padding: 20px 0;
   z-index: 100;
+  height: calc(100vh - 230px);
```
rsimon, line 31: I can see that breaking in various ways. (Browser differences, people having different font sizes, etc.) Is this necessary? I can help clean up the CSS later. But can't test locally right now (the staging backend doesn't have the  `projects.is_locked` column yet).

src/apps/dashboard-projects/List/ProjectEntry.tsx
```diff
@@ -2,7 +2,7 @@ import { useState } from 'react';
 import { joinProject } from '@backend/helpers';
 import { supabase } from '@backend/supabaseBrowserClient';
 import { Avatar } from '@components/Avatar';
-import { OwnerPill } from '@components/OwnerPill';
+import { LockedPill } from '@components/LockedPill';
```
rsimon, line 5: So we no longer use the "owner" pill in the list view? Or are we deprecating the owner pill altogether?

src/components/Filter/HeaderFilterAction.css
rsimon, line 1: If this component isn't used elsewhere, it might be worth putting it as a subcomponent in `src/apps/dashboard-projects`. If not, we should IMO give it a more explicit name than "Filter", to make it more obvious what this is for.

src/components/Sort/HeaderSortAction.css
rsimon, line 1: Same as with `Filter` above: might be worth putting it into `src/apps/dashboard-projects` if it's only used there. Or give it a more explicit name?

rsimon (approved):
Looks good! I added DE labels and made a few suggestions. But those are really just optional. (P.S.: I couldn't actually test this in practice, because the staging backend doesn't have the `is_locked` column yet. Let me know when that's deployed, and I'll start working on the necessary changes to the annotation view.)

src/apps/dashboard-projects/List/ProjectEntry.tsx
lwjameson: Yes. I believe we decided that since all projects on the "My Projects" page are ones you own it did not make much sense.

src/layouts/BaseLayout.astro
```diff
@@ -17,12 +17,15 @@ const backgroundColor = config.branding.background_color;
 const favicon = `/${config.branding.favicon}` || '/favicon.svg';
 ---
 
-<!DOCTYPE html>
+<!doctype html>
```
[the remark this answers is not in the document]
lwjameson: https://developer.mozilla.org/en-US/docs/Glossary/Doctype

Apparently this is now the standard. It was auto-corrected.

src/components/Filter/HeaderFilterAction.css
lwjameson: Fixed.

src/components/Sort/HeaderSortAction.css
lwjameson: Fixed

src/components/Filter/HeaderFilterAction.css
rsimon: Great, thanks!

rsimon:
Looks great now! (I added one small refactoring [here](https://github.com/recogito/recogito-client/pull/264/commits/faa976074b5371fc8c465537606e87666ae59527). The `SortFunction` was still in the `components/Sort` folder, and the `index.ts` in the sort folder re-exported the SortAction component from `../../apps/dashboard-projects/Header/Sort/HeaderSortAction`.)

rsimon (approved):

PR #268, opened 2024-10-03
Bugfix: Broken CSV Exports for Baselayer Context

rsimon:
## In this PR

This small PR fixes a bug reported by @jsmueller7 in our last meeting: the CSV export for the Baselayer Assignment returned an error. Turns out this was simply a matter of how the download filename got generated from the context name (which is `null` for the base context).


lwjameson (approved):
Looks good.

PR #269, opened 2024-10-03
Attempted fix for SAML Redirect Flow Issue

rsimon:
## In this PR

In our last meeting @jsmueller7 reported issues with the login redirect flow: existing redirects weren't properly removed from localStorage, leading to the same redirects being applied later when they shouldn't. @jsmueller7 later reported that this is only happening for SAML.

I _think_ I have an idea why this might be happening, but cannot test this myself with SAML. @lwjameson: can you take a look at these proposed changes? I think it's simply a matter of clearing out localStorage after the SAML flow returns?


lwjameson (approved):
That will work.

Issue #266, opened 2024-10-02
Filtering options aren't dynamic


jsmueller7:
When I open a document in the base layer, I can filter by “Created by” and “Tags”. In my opinion, the suggestions for “Created by” and “Tags” should depend dynamically on the selected layers. The current status is that I can filter on a tag or a person even if they do not appear in the selected layer.

rsimon:
Agreed! I'm preparing a PR with some smaller annotation view fixes, which will include this change.

PR #270, opened 2024-10-03
Annotation View Fixes

rsimon:
## In this PR

Two tiny fixes to the annotation view:
- Issue #266: __Creator__ and __Tag__ filters are now aware of the current layer filter.
- The IIIF thumbnail sidebar flickered, while caught in an endless loop showing & hiding the horizontal scrollbars. (See below.) This did not happen on all screen resolutions and browser window sizes. Frankly, I wasn't able to fully diagnose the issue. But it must be related to the JS-based sizing that's done by the virtual window library we're using to render only the visible portion of the thumbnail strip.

https://github.com/user-attachments/assets/8a3c39e6-850b-4d5a-b33c-f99034c92cd0


lwjameson (approved):
Looks good.

PR #271, opened 2024-10-03
Fix for availableLayers not updating when assignment is deleted

lwjameson:
## In this PR

When assignments are deleted, new assignments who the deleted assignment as an available layer. If that is chosen, the assignment is broken.  This PR moves the setAssignments call, which updates the availableLayers state, after the delete. This does require eliminating the previous optimistic update of assignments, but seems the safest and most expedient approach.


rsimon (commented):
I must admit I don't quite understand why this wasn't working before. Is this a problem just for the time period between deleting and the `archiveAssignment` call completing?

rsimon (approved):

PR #274, opened 2024-10-04
uppy Library Update

rsimon:
This PR updates the `@uppy` library dependencies we use to implement file upload to the latest versions. 

Previously, we had an issue that uppy returned [undefined responses in this line](https://github.com/recogito/recogito-client/blob/develop/src/backend/storage/iiif.ts#L61). This seems to have been fixed in the mean time, which means we can now upgrade again. Note that meanwhile, `@uppy` has a new major release out (v4). There were a few minor breaking changes to address.


lwjameson (approved):
Looks fine.

PR #275, opened 2024-10-04
Deep Linking

rsimon:
## In this PR

This PR adds support for "deep linking" of annotations.
- The currently selected annotation gets represented in the page URL's hash fragment (`selected=[uuid]`), giving each annotation a unique URL that users can copy and paste.
- The selection state is restored from the hash in the following two situations:
   - When the page is initially loaded.
   - When the `popstate` event occurs (= browser back button).

Furthermore, there are two additions to the annotation editor component:
- A new "copy link to annotation" option in the context menu.
- Formatting logic for Quill, which detects an annotation URL and replaces is with a clickable "shortcode" link.

### IIIF Multi-Page Presence

This PR also adds a simple version of "multi-page presence" to the IIIF view. When users are in the same multi-page IIIF image, but on a different page, the thumbnail now shows a small dot in that user's color, to indicate their presence on the page.

### Known Issues

- ~~__Dangling PDF popup__. When linking to a PDF annotation that's several pages down the document, the annotation popup is "dangling" at a default top-left position, while the document scrolls down. Reason: although the annotation exist, and is (technically) selected, its visual representation is not yet rendered because of PDFjs's lazy rendering. I'm trying to resolve this in the `@recogito/react-pdf-annotator` lib.~~ __Update: fixed__

- __Initial loading of the first IIIF image__. When deep-linking to a IIIF annotation, the annotation view will __always__ load the first page initially, and then jump to the appropriate target page. This happens quickly, but introduces a "flicker" for the user where the first page is briefly visible before the actual target page comes up. (Technical reason: we are creating an OpenSeadragon viewer first; then an Annotorious instance on that viewer; and then let Annotorious handle annotation fetching. If we want to avoid loading the first page, we'd need to lift the annotation fetching out of Annotorious. This would be a bigger change. I'd recommend not doing this for now.)

- __Inaccessible links__. An annotation link can point to a document which the current user may not have access privileges to. The current implementation __does not check or validate this__. In theory, we could add logic that attempts to visit the link when the popup opens, and disables it in case it gets back a non-HTTP 200 response.


lwjameson (approved):
Monster PR! Looks good.

PR #276, opened 2024-10-04
Fix for logout

lwjameson:
## In this PR

- An issue emerged when we logged out. I believe it is a regression issue from a new Supabase version. We are getting a 403  return from logout and you sit there on the frown face screen.

https://github.com/supabase/supabase-js/issues/1170

It may be that we may not want to ever care if logout fails? At least from a routing standpoint


rsimon (approved):
Looks good!

PR #277, opened 2024-10-07
Better messaging on team assignment wizard

lwjameson:
#Summary

Users have been confused about assignment teams and project admins.  This PR makes it more explicit that Project Admins are automatically members of all assignments.


rsimon (approved):
Looks good! I added DE labels.

PR #287, opened 2024-10-10
Locked Projects: Annotation View

rsimon:
## In this PR

This PR adds the necessary UI plumbing to make the annotation views appear "read-only" to the user for locked projects.

The following changes happen in the UI if the project is locked:
* Text:
  * Text selection no longer triggers annotation. Native browser selection as such remains possible, but is styled in a "disabled" style, to signal the change to the user.
* Image:
   * Drawing is disabled.
   * Selecting an annotation will no longer make the target editable, regardless of privileges.
* Annotation popup: 
   * No more reply field.
   * No more "..." actions menu button. (Note: we may want to change this after merging the [deep-linking PR](https://github.com/recogito/recogito-client/pull/275), so that the option remains to copy the annotation link.)
* Annotation sidebar: 
   * No more reply field on selected annotation.
   * No more "..." button on annotation cards.
* Toolbar: 
   * No more privacy mode selector.
   * No delete button when an annotation is selected.
   * No more drawing tools (image view).
* Notes sidebar
  * No more "New Note" button.
  * No more reply field when selecting a Note.

__Note:__ I branched this PR off the `lwj/locked-projects` branch, so you can see my additions more easily, @lwjameson. I guess we can either merge the `lwj/locked-projects` into `develop` first, or we merge this branch into `lwj/locked-projects`.


lwjameson (approved):
Looks good. I will merge my PR first.

PR #288, opened 2024-10-10
Filter By Contributor

rsimon:
## In this PR

This PR implements a recently requested change in the annotation filter panel: instead of filtering by the __creator__ of an annotation, the filter panel now filters by __contributor__. This way, annotations are also shown if a particular user has replied to the annotation.


lwjameson (approved):
Looks good.

Issue #289, opened 2024-10-14
Unable to comment with link


jsmueller7:
I've just tried with client version `1.1.2`: I've not been able to comment or reply using the link insertion function. Also tested on the staging server.

rsimon:
Hi @jsmueller7,

can you describe what happens? Does the popup close as soon as you click the link button? (If so: that sounds like a regression in the base text annotator library - which however should be fixed in v1.1.2, so it might be a deployment problem, with a stuck library version?)

If not: does it happen in text (PDF vs. txt/TEI) or in both interfaces (text + image). Also, can you confirm which staging system you are testing on? (I can't seem to reproduce this on ours at http://recogito-staging.covecollective.org.)

jsmueller7:
Hi @rsimon,

when I mark a text section and klick the link symbol, the popup "Link" opens. After pasting an URL and sending I return to the regular - but empty - annotation editor. This happens with all text formats (I haven't tested on images). It also happens on the PS staging server. That's why I believe it could be a general problem.

I've just tested with another approach: I inserted the URL and then I added some whitespaces and removed them - this way I was able to create a link comment. Unfortunatelly I haven't been able to succeed with this workaround twice... Still trying to find a pattern.

jsmueller7:
The one time I was able to create the link I was on the PS staging server. Still I can't reproduce.
In my german UI also the `Link` label went missing, just like with `e-mail` in the invitation pop up.

jsmueller7:
Allright. Now I've been able to reproduce the workaround: I simply need to insert the link twice, without leaving the annotation editor inbetween. So the workaround at the moment is:

* Mark a text or image section,
* Open the link editor,
* paste an URL,
* return to the empty annotation editor,
* again start the link insertion,
* now you'll have a hyperlink.

rsimon:
Ah, ok I think I know what's going on. The feature "works as designed", in a sense (but isn't really designed right...)

The link button is supposed to add a link to an existing text snippet. It will only work if you have text selected in the comment box. I.e. let's say:
- You write the text "Here's a link to example.com"
- You select "example.com" and click the link button
- You paste the link "https://www.example.com" into the URL popup and click "Save"
- The comment now shows "Here's a link to [example.com]" - with "example.com" as a clickable link

[...]

PR #290, opened 2024-10-14
Comments: Pasting Links + Editing Behavior

rsimon:
## In this PR

This PR addresses issues reported in #289:

* Previously, it was not possible to paste a link into a comment if no text was selected. After this PR, a link pasted without an active selection will paste the link URL as a clickable text, at the current cursor position.
* The "Link" UI label was missing for DE.
* The dialog did not have description text. This caused Radix to fire a warning about accessibility. There is now dialog description for the link dialog, similar to the dialogs for image and video.

Separately, this PR adds another fix: when re-opening a comment for editing, the Quill RTE no longer showed the toolbar. See below. This PR adds the toolbar, also in the "edit comment" card state. 

__Before:__
Bildschirmfoto 2024-10-14 um 15 03 07

__After:__
Bildschirmfoto 2024-10-14 um 15 03 23


rsimon:
__Update:__ I added a small cosmetic change. When the an annotation is created, it abruptly swaps from 'new annotation' to selected annotation state in the sidebar. I added a view transition which smooths out the jump. This is only supported on modern browsers (which support the view transition API). But won't cause any problems on older browsers. (Those simply have to live with the jump.)

lwjameson (approved):
Looks good!

PR #291, opened 2024-10-17
Exporting info.json URL to CSV

rsimon:
## In this PR

When a CSV export includes image annotations, the URL for the images' `info.json` is now exported in a new column.

Note that ideally, we'd want to export the __canvas ID__ rather than the __info.json URL__. However that is not currently available in the annotation. Perhaps something we may need to revisit as part of the planned [IIIF export](https://github.com/performant-software/vico/issues/313) feature.


lwjameson (approved):
Looks good.

PR #293, opened 2024-10-18
Change to messaging for deleting a document from the Doc Library

lwjameson:
## In this PR

We had a bug found on one of our instances when a user deleted a private document from the document library that was being used by a project.  I have fixed the server code that allowed it, and this PR changes the messaging on a failed delete which may inform the user as to why.

** Will need the german updated.


rsimon (approved):
DE label added.

PR #86, opened 2023-11-16
Fixes #214

rsimon:
Addresses double-rendering of comments in annotation popup: https://github.com/performant-software/vico/issues/214

PR #294, opened 2024-10-21
#86 - Document Default View

dleadbetter:
This pull request adds a new setting to the Project Settings page to allow project owners to set the default view for the right panel when a document is opened. By default, the right panel will remain closed, but project owners can elect to have the right panel open to the annotations or notes tabs.

A hash parameter called `rtab` will be passed to both the `ImageAnnotationDesktop` and `TextAnnotationDesktop` components to determine how the right panel should be displayed. Valid values for the hash parameter are `closed`, `annotations`, and `notes`.

## Project Settings
Screenshot 2024-10-21 at 10 19 07 AM

## Default View
Screenshot 2024-10-21 at 10 19 32 AM


src/apps/annotation-image/ImageAnnotationDesktop.tsx
```diff
@@ -87,7 +89,7 @@ export const ImageAnnotationDesktop = (props: ImageAnnotationProps) => {
 
   const [leftPanelOpen, setLeftPanelOpen] = useState(false);
 
-  const [rightPanelOpen, setRightPanelOpen] = useState(false);
+  const { rightPanelOpen, rightPanelTab, setRightPanelOpen } = useAnnotationsView();
```
rsimon, line 92: I think that's a good hook to have - with one caveat below...

(BTW, I might lean towards renaming it `useAnnotationViewState` or `useAnnotationViewUIState`. Not sure what you think?)

src/apps/annotation-image/RightDrawer/RightDrawer.tsx
```diff
@@ -51,7 +52,7 @@ export const RightDrawer = (props: RightDrawerProps) => {
     }
   });
 
-  const [tab, setTab] = useState<'ANNOTATIONS' | 'NOTES'>('ANNOTATIONS');
+  const [tab, setTab] = useState<'ANNOTATIONS' | 'NOTES'>(props.defaultTab);
```
rsimon, line 55: This is now __initializing__ the state from the hash param. But it's not keeping the hash in sync for state changes. Ideally, I think we'd like to have some kind of hook that acts as a drop-in replacement for `useState`, but works through the hash. That would give us share-able URLs on the one hand, but would use be useful for another purpose. Some context:

The root components (`ImageAnnotationDesktop`, `TextAnnotationDesktop`) need to be aware of whether the right tab is open, and whether it shows the in 'ANNOTATION' tab. It needs this in order to decide whether it should use floating popups when annotations are selected.

Right now, the implementation is a spaghetti mess. Right-panel closed/open state is maintained at the top; tab change triggers a callback, which then updates another state var `usePopup`. 

Now we could build everything off this nice & new `useAnnotationViewUIState` hook, which we'd keep at the top, and could use as the source of truth for the popup. (And any other things that might come up in the future.) __However__, in order for this to work we'd need to keep this state in sync with the UI. I wonder what your thoughts are @dleadbetter and @lwjameson?

rsimon (changes requested):

src/apps/annotation-image/RightDrawer/RightDrawer.tsx
lwjameson: @rsimon I like it.  We should take the opportunity to future-proof when we can, and I think this might simplify in the end, if only simplifying the ability to understand the code. Always good in Open Source.

rsimon (approved):
Thanks @dleadbetter, looks great!

lwjameson (approved):
Looks good.

PR #297, opened 2024-10-24
Adjust Invite User Flow

lwjameson:
## In this PR

With the switch over to using the Supabase PKCE Auth flow, some adjustments had to be made. 

- The PKCE flow does not support invite user, so in order to keep the same UX the flow needed to be changed.
  - Invite User now uses an API endpoint which creates the user and sends them an email instead of using the Auth inviteUserByEmail functionality.
  - This endpoint creates an encrypted token which includes the user ID as well as the epoch time that it was generated.
  - The user, when clicking the 'Accept Invite' button is sent to a new client page.
  - This page is similar to the reset password page, but sends it's response to a new API endpoint which validates the token and sets their password.


src/apps/accept-org-invite/AcceptOrgInvite.tsx
```diff
@@ -0,0 +1,134 @@
...
+        } else {
+          setSuccess(true);
+        }
+        setBusy(false);
+      });
+    }
+  };
+
+  return (
+    <div className='reset-password'>
```
rsimon, line 66: That should probably be changed to something like `accept-org-invite`.

template.html
```diff
@@ -0,0 +1,385 @@
...
+              border="0"
+              cellpadding="0"
+              cellspacing="0"
+              class="main"
+            >
+              <!-- START MAIN CONTENT AREA -->
+              <tr>
+                <td class="wrapper">
+                  <img src="{{ .SiteURL }}/img/branding/email/top-logo.svg" />
+                  <p>Hello from Cove Studio</p>
```
rsimon, line 327: Is this a Cove-specific template? Or something that also needs branding?

src/pages/api/accept-new-user.ts
```diff
@@ -0,0 +1,89 @@
+import { createClient } from '@supabase/supabase-js';
+import type { APIRoute } from 'astro';
+import { decrypt } from '@backend/crypto';
+import type { ApiAcceptOrgInvite } from '../../Types';
+import { error } from 'node_modules/astro/dist/core/logger/core';
```
rsimon, line 5: That's probably a leftover from a wrong autocomplete?

src/pages/api/invite-user.ts
```diff
@@ -66,13 +85,157 @@ export const POST: APIRoute = async ({ request, cookies }) => {
...
+  const mailOptions = {
+    from: MAIL_FROM_ADDRESS,
+    to: body.email.toLowerCase(),
+    subject: t['Join Recogito'],
+    html: html,
+  };
+
+  const sendResp = await transporter.sendMail(mailOptions);
+
+  console.log('Message sent: ', sendResp.messageId);
```
rsimon, line 225: Leftover debug log :-)

src/pages/api/invite-user.ts
```diff
@@ -66,13 +85,157 @@ export const POST: APIRoute = async ({ request, cookies }) => {
...
+          fontWeight: 'normal',
+          padding: {
+            top: 0,
+            bottom: 16,
+            right: 24,
+            left: 24,
+          },
+        },
+        props: {
+          text: t['_join_recogito_message_'].replaceAll(
```
rsimon, line 174: Changed that to `replaceAll` since I ended up with a DE version that uses the instance name twice.

rsimon:
Nice!

rsimon (approved):
Nice!

template.html
lwjameson: This is some helper files.  I will be moving them.

PR #303, opened 2024-11-05
Annotation View: Active Layer + Color Coding

rsimon:
## In this PR

This PR addresses two issues in the annotation view, raised in this week's call with the client:
- __Active layer:__ if no active layer is present in the current context/document, the annotation view used to silently fall back to the first available layer. Since this should never happen, the fallback was removed. Additionally, a check was introduced that verifies that: i) there are >0 layers, ii) there is exactly one active layer. If the check does not pass, a fatal error is thrown with a user-readable message.
- __Persistence for the color-coding setting:__ as requested by the client, the color-coding setting is now persisted to local storage. (Persistence is specific to the current context/document.)


lwjameson (approved):
Looks good!

PR #304, opened 2024-11-08
Invite Users CSV fix

lwjameson:
## In this PR

- addresses: https://github.com/performant-software/vico/issues/335


rsimon (commented):
So... this is an odd one. Given the error message, I'd say papaparse was indeed able to parse the file. The message sounds more like a warning ("defaulted to ','"). So I wonder if, instead, we're using papaparse wrong in this case.

- I don't think we should force a comma delimiter. This by-passes papaparse's auto-detection, which would otherwise be useful.
- Given that `result.errors` seems to include warning messages even if parsing was successful, we should probably change to something like this instead?

```
if ((result.data || []).length > 0) {
  setCsv(parseUsers(result.data) || undefined);
} else if (errors.length > 0) {
  const errors = result.errors.map(e => e.message).join('\n');
    props.onError(errors);
}
```

SteveMarvin:
Yay!

PR #306, opened 2024-11-08
Fix sending reset email twice

lwjameson:
## Summary

- Addresses: https://github.com/performant-software/vico/issues/331
- https://github.com/performant-software/vico/issues/328

### DIFF
--- a/src/layouts/BaseLayout.astro
+++ b/src/layouts/BaseLayout.astro
@@ -15,6 +15,16 @@ const textColor = config.branding.contrast_color;
 const backgroundColor = config.branding.background_color;
 
 const favicon = `/${config.branding.favicon}` || '/favicon.svg';
+
+function truncateString(str: string, maxLength: number) {
+  if (str.length > maxLength) {
+    return str.substring(0, maxLength) + '...';
+  } else {
+    return str;
+  }
+}
+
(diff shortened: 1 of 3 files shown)


src/layouts/BaseLayout.astro
```diff
@@ -15,6 +15,16 @@ const textColor = config.branding.contrast_color;
 const backgroundColor = config.branding.background_color;
 
 const favicon = `/${config.branding.favicon}` || '/favicon.svg';
+
+function truncateString(str: string, maxLength: number) {
+  if (str.length > maxLength) {
+    return str.substring(0, maxLength) + '...';
```
rsimon, line 21: I wonder if that's necessary? Browser tab titles will truncated anyway. We may just want to ensure we're not hitting some spec-defined limit. (Not sure if one exists, but we could e.g. cut at 256 or 1024 chars?)

rsimon (approved):
Looks good!

src/layouts/BaseLayout.astro
lwjameson: Based on this:

https://www.w3.org/Provider/Style/TITLE.html

I will limit the length to 64 (actually 61 plus the '...')

Issue #281, opened 2024-10-09
Header bar doesen't break well on small screens


jsmueller7:
First I tested this on my smartphone. But this seems also to be a problem on tablets and you can even reproduce this in a regular browser window: If you move the window together, a scroll bar appears at the bottom. At the top right, the toolbar protrudes.

![Bildschirmfoto vom 2024-10-09 14-00-11](https://github.com/user-attachments/assets/7bf0b60d-fd06-4e3b-9737-47fb0f1224b3)

SteveMarvin:
This issue is approved, but I cannot move it. @rsimon @lwjameson

PR #307, opened 2024-11-13
Responsive Annotation Toolbar

rsimon:
## In this PR

Addresses issue reported in #281 and [here](https://github.com/performant-software/cove-recogito/issues/124): the toolbar is now more responsive, and adapts to smaller screens. The current approach should work well down to 640px width.

### 1. Truncated Title

There is now a maximum width enforced for the assignment- and document titles. Labels are clipped with ellipses if they exceed.

### 2. Dynamic Removal of Toolbar Items

On initial load and each time the window gets resized, if the toolbar overflows, a fixed set of toolbar items gets removed. Instead, a three-dot "more" button is shown. Also, the toolbar switches from a three-group (left, center, right) to a two-group (left, right) layout. (Note that there is only a single breakpoint. Given we don't have many menu items, I don't think it makes sense to collapse in multiple stages.)

In __text mode__, the following items are removed:
- PDF zoom controls
- Color coding selector and legend

__Full width:__
Bildschirmfoto 2024-11-13 um 14 55 23

__Collapsed:__
Bildschirmfoto 2024-11-13 um 14 55 35

In __image mode__, the following items are removed:
- Privacy selector
- Color coding selector and legend

__Full width:__
Bildschirmfoto 2024-11-13 um 14 54 44

__Collapsed:__
Bildschirmfoto 2024-11-13 um 14 55 04

When clicking the "more" menu button, removed items are available in an additional popover panel.

Bildschirmfoto 2024-11-13 um 14 56 50


lwjameson (approved):
Nice!

PR #310, opened 2024-11-15
Assignment layers fix

lwjameson:
# Summary

 - Addresses: https://github.com/performant-software/vico/issues/341
 - Fixes some accessibility warnings

Documents added to the base layer were not updating the availableLayers value, so a refresh was required to see base layers on initial project creation.


rsimon (approved):
Looks good! I Added the DE label.

PR #314, opened 2024-11-21
"Copy Annotation Link" menu option

rsimon:
## In this PR

This PR should address issue https://github.com/performant-software/vico/issues/349.

Some extra notes, because I believe the issue description isn't quite accurate.
- There was __no difference in behavior between popup vs. sidebar__. The issue was the same for both cases.
- There was no problem with __edit/delete options__. At least I could not reproduce a problem.

However, what __was__ broken is this:
- The three-dot context menu was only shown if you had edit permissions (creator/admin) on the annotation. Therefore,  you would not be able to access the "copy link" option - unless you were creator/admin.
- The "copy link" option was shown on all context menus, including those on replies. (Whereas we only want to show it in the top-most context menu, alongside the first comment.)

This PR makes the following fixes:
- The three-dot context menu is now __always__ shown. (Exception: read-only annotations. Probably less important, and we don't have a design for this. Right now, the "layer" icon occupies the space where otherwise the three-dot menu would be.)
- If you don't have edit permissions on the annotation, there is only a single menu option: "copy link". Creator/admins still get all options (copy link, edit comment, delete annotation).
- The "copy link" option is no longer shown in the reply-menus.

### DIFF
--- a/src/components/Annotation/AnnotationCardSection.tsx
+++ b/src/components/Annotation/AnnotationCardSection.tsx
@@ -240,34 +240,35 @@ export const AnnotationCardSection = (props: AnnotationCardSectionProps) => {
           {editable ? (
             <QuillEditorToolbar
               i18n={props.i18n} />
-          ) : canEdit ? (
+          ) : isPrivate ? (
             <div className="annotation-header-right">
-              {isPrivate ? (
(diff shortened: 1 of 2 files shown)


lwjameson (approved):
Looks good!

PR #317, opened 2024-11-22
Zombie Annotations

rsimon:
## In this PR

This PR fixes another loophole that lead to zombie annotations (= annotations without tag/comment that should have been auto-deleted, but weren't). The fix is in the `@recogito/annotorious-supabase` package ([see here](https://github.com/recogito/annotorious-supabase/commit/5df86ac9c36f76d3ca7dde69fceed8cf0e787383)). This PR only updates the version.

The cause was a race condition, that occurred when the user deleted an annotation __instantly__ after creating it. In this case, the local deletion happened before the remote 'create' Postgres events returned, causing the annotation to get re-generated. Full story is in this issue: https://github.com/performant-software/vico/issues/351

P.S.: the PR also updates a few other minor libraries & fixes a deprecation warning.


src/apps/annotation-text/AnnotatedText/AnnotatedText.tsx
```diff
@@ -146,7 +146,7 @@ export const AnnotatedText = (props: AnnotatedTextProps) => {
           )}
 
           {props.usePopup && (
-            <TextAnnotatorPopup
+            <TextAnnotationPopup
```
rsimon, line 149: Fixes deprecation warning

src/components/AnnotationDesktop/UndoStack/UndoStack.tsx
```diff
@@ -21,9 +21,9 @@ export const UndoStack = (props: UndoStackProps) => {
   const { selected } = useSelection();
 
   const deleteIfEmpty = (annotation: Annotation) => {
-    const currentState = store.getAnnotation(annotation.id);
+    const currentState = store!.getAnnotation(annotation.id);
```
rsimon, line 24: Fixes TS warnings

lwjameson (approved):
Looks good!

Issue #313, opened 2024-11-20
Assignment titles trend to exceed the given space


jsmueller7:
In practice, lecturers tend to assign longer titles than the UI allows for. This leads to a somewhat messed up look.

![Bildschirmfoto vom 2024-11-20 13-47-38](https://github.com/user-attachments/assets/aa1eab26-432c-4df2-80b4-8a3873925eae)

PR #321, opened 2024-11-25
Updates to collaboration screen

lwjameson:
# Summary

- Addresses the styling and functional issues in https://github.com/performant-software/vico/issues/348
- Addresses https://github.com/recogito/recogito-client/issues/313


src/backend/crud/projects.ts
```diff
@@ -8,7 +8,7 @@ import type {
 import type { Response } from '@backend/Types';
 import { getUser } from '@backend/auth';
 import type { ApiPostInviteUserToProject } from 'src/Types';
-import type { InviteListEntry } from '@apps/project-collaboration/InviteListOfUsers/InviteListOfUsers';
+import type { InviteListEntry } from '@apps/project-collaboration/InviteListOfUsersDialog/InviteListOfUsers';
```
rsimon, line 11: That file doesn't seem to exist any more. (Should be `InviteListOfUsersDialog`?)

In general, I think I'd bundle the types with the backend code, and then import from there into the UI code, rather than vice versa. But that's really just a matter of taste I guess & no strong opinion either way.

rsimon (approved):
Looks good! (I added the DE label.)

PR #322, opened 2024-11-26
TEI Export Format Fix

rsimon:
## In this PR

This PR fixes the markup format for TEI annotation exports. Previously, the `<standOff>` element was placed as a child of the `<teiHeader>`, which was not standard-compliant.

This PR places `<standOff>` elements directly at the root, as a sibling to the `<teiHeader>` and `<text>` elements.

- If there is no `<teiHeader>`, and no other existing `<standOff>` elements, the new `<standOff>` element is injected as the first child of the root `<TEI>` element.
- Otherwise, the new `<standOff>` element is injected after the `<teiHeader>` and the last existing `<standOff>` element. (I.e. typically between the last existing `<standOff>` and the `<text>`.)
- A `type` attribute is added like so: `<standOff type="recogito_studio_annotations">...</standOff>` 

__Note:__ I realized that there was a bug with "Notes" (= annotations without a text selection). Notes have so far broken the export! For now, I patched this in the following way:
- Notes are exported as TEI annotations without a target.
- Notes that are embedded in an uploaded TEI are __not__ (yet) re-generated as Notes in the Recogito interface! (I made some first steps to this. But it will take a bit more work.)


rsimon:
Hi @lwjameson, I just added support for displaying embedded notes. (In [this commit](https://github.com/recogito/recogito-client/pull/322/commits/1dde2a51fc172cd2a078502ae219e73135a9a1bd).)

> Would any previously exported annotated TEI be broken if used in Recogito?

Good point. After this PR, `<standOff>` elements inside the `<teiHeader>` would simply be ignored. I.e. while previously exported TEI documents would break on import, the annotations would not be shown. I could easily change that, however, and accept `<standOff>` elements at any place in the TEI document.

PR #326, opened 2024-12-04
"Notes" Sidebar Bugfixes

rsimon:
### In this PR

This PR fixes a set of bugs discovered recently on the Pelagios instance.

With this PR:
- The Notes list no longer displays duplicates.
- Sorting the Notes list no longer causes a crash if the document includes embedded notes.
- If the initial page configuration is set to `rtab=notes`, the interface no longer crashes. (This was due to a Nullpointer caused by a race condition, where the Notes list was rendered before the annotation interface was fully initialized.)


lwjameson (approved):
Looks good

PR #330, opened 2024-12-05
More Robust Handling for IIIF info.json URLs

rsimon:
## In this PR

This PR makes resolution of IIIF `info.json` URLs more robust. The test case for this was the following manifest, which didn't properly open previously:
https://iiif.quartexcollections.com/digitalcollections-baylor/iiif/cfa534b4-3784-4c64-ba79-20706bd9ab28/manifest


lwjameson (approved):
Looks good!

PR #332, opened 2024-12-06
Fix for SSO and Keycloak redirects

lwjameson:
# Summary

With the PKCE implementation, the old redirect methods no longer work for SSO and Keycloak logins.  We now have to use the redirect options for calling the server-side key verifications. Since clients are redirected to `/{lang}/projects` by default, this PR adds a check on that route as to whether there is an active redirects and then executes the redirect.


rsimon (approved):
Look good!

PR #83, opened 2023-11-14
Updating the client - docker - instructions

lwjameson:
This pull request adds some instructions for manually updating the recogito client via docker.

PR #333, opened 2024-12-06
#83 - Image collections

dleadbetter:
This pull request refactors the `MetadataModal` into the `EditMetadataModal` and `ViewMetadataModal` components.

## EditMetadataModal
This component will function basically the same as the previous `MetadataModal`. Users have the ability to modify a finite list of metadata attributes. This modal will only be available to the creator of the document. All other users will see the `ViewMetadataModal`. The `readOnly` prop as been removed.

## ViewMetadataModal
This component will render the key/value pairs contained in the `meta` attribute of the `documents.metadata` field. The values will differ between documents uploaded by users and documents added as part of the COVE image gallery.

Screenshot 2024-12-06 at 11 23 14 AM
Screenshot 2024-12-06 at 11 23 33 AM
Screenshot 2024-12-06 at 11 54 32 AM
Screenshot 2024-12-06 at 11 54 37 AM


src/i18n/en/project-home.json
```diff
@@ -7,7 +7,7 @@
   "Clear": "Clear",
```
rsimon, line 7: I added DE labels.

Just an additional note: the DE labels are not automatically filled in, if only the EN labels exist. I.e. if there are no DE labels, UI elements will be blank in the DE interface. (We should always fill the DE labels, even if we temporarily only fill them with the English versions.)

rsimon (approved):
Looks good!

PR #339, opened 2024-12-18
TEI Dowload: Exporting Tags as Taxonomy Element

rsimon:
## In this PR

This PR changes the way tags are exported in the TEI download. 

__The problem:__ for each annotation, tags used to be exported as an `<rs>` element with an `@ana` attribute that simply joined the tag labels. Therefore, tags that included whitespace would break the export:

`<rs ana="TagA TagB My Tag With Whitespace"/>`

__This PR makes the following changes:__

- A `taxonomy` element is created in the `teiHeader`, which lists all distinct tags in the document, and assigns them (random) IDs.
- The `@ana` attribute now references the IDs instead of the tag labels.
- The `@ana` attribute is now on the annotation directly, not on an extra `rs` element.

__Additional Notes:__

- If there is no `teiHeader` in the document, a header is created and inserted automatically.
- If there are existing `taxonomy` elements in the document, a new `taxonomy` is appended after them.
- The new taxonomy gets a (random) `xml:id` assigned during export. The export verifies that this ID is unique, and not used by another `taxonomy` in the document yet.
- Tag ids have the form `{taxonomy-id}.{random-tag-id}`.
- In the annotation view, when annotations are parsed from the embedded layers, their tag labels are resolved against the taxonomies in the document.
- A new rule was added to the CSS stylesheet to hide TEI `taxonomy` elements in the annotation view.

### DIFF
--- a/src/apps/annotation-text/LeftDrawer/LeftDrawer.css
+++ b/src/apps/annotation-text/LeftDrawer/LeftDrawer.css
@@ -2,6 +2,7 @@
   flex-grow: 0;   
   flex-shrink: 0; 
   flex-basis: 0;
+  overflow-x: hidden;
   position: relative;
   scroll-padding: 40px;
   z-index: 9;
@@ -11,5 +12,5 @@
   position: absolute;
   top: 5px;
   left: 0;
-  width: 240px;
+  width: 280px;
 }
--- a/src/components/Annotation/TagList.css
+++ b/src/components/Annotation/TagList.css
@@ -9,6 +9,10 @@
   padding: 0 0 10px 0; 
 }
 
+.annotation .taglist button {
+  height: auto;
+}
+
 .annotation .taglist.editable {
   padding-bottom: 0;
 }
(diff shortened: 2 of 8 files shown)


src/components/Annotation/TagList.css
```diff
@@ -9,6 +9,10 @@
   padding: 0 0 10px 0; 
 }
 
+.annotation .taglist button {
+  height: auto;
```
rsimon, line 13: I noticed that tags appeared to high in the annotation editor. This seems to have been caused by a blanket `height: 38px` rule that got introduced for the `button` element at some point. This fix patches the problem.

src/apps/annotation-text/LeftDrawer/LeftDrawer.css
```diff
@@ -11,5 +12,5 @@
   position: absolute;
   top: 5px;
   left: 0;
-  width: 240px;
+  width: 280px;
```
rsimon, line 15: Just minor style tweaks to the side panel.

lwjameson (approved):
Looks good!

PR #346, opened 2025-01-03
Lwj/add all members to assignment

lwjameson:
## In this PR

- Add the ability to add all team members to a given assignment.
  - All team members are added at the time of assignment creation
  - People who join the team after assignment creation are added automatically
  - If it is turned off, all team current team members will still ge on the team, but future members will not be automatically added.


rsimon (approved):
Looks good! I added DE labels.

PR #347, opened 2025-01-06
#139 - Document metadata

dleadbetter:
This pull request adds the ability to add arbitrary metadata to `documents` records. It also provides the ability to view/modify document metadata from within the annotations view.

## View/Edit Metadata Modal
From the projects documents view, users can view or modify document metadata. Users can modify the metadata is they are the owner of the document.

Screenshot 2025-01-06 at 2 43 02 PM

## View/Edit Metadata Modal
New key/value pairs can be added by clicking the "Add Metadata" button.

Screenshot 2025-01-06 at 2 43 14 PM

## View/Edit Metadata Modal
The document name is also edit via the modal.

Screenshot 2025-01-06 at 2 43 20 PM

## Annotations View
Document metadata is also able to be viewed/edited via the annotations view. Internal document refers to the metadata stored in the `documents` record.
Screenshot 2025-01-06 at 2 43 36 PM

## Annotations View
For a document added via IIIF manifest, the metadata from the manifest is also visible under the "External" heading. This metadata is never editable.

Screenshot 2025-01-06 at 2 43 58 PM

## Annotations View
For documents with no metadata, a simple message displays on the "Metadata" tab.
Screenshot 2025-01-06 at 2 44 10 PM


rsimon (approved):
This is great!

The only small issue I found was in the image annotation interface, with the embedded IIIF metadata. I looked at a few manifests, and some included URLs as field values. Not sure if we want to make those clickable? But they definitely have an overflow problem at the moment–see screenshot below. We might want to stack the field headings above the field values perhaps, to make more space; or apply a `flex-shrink: 0` to the label and an `overflow:hidden; text-overflow: ellipsis` to the value, or something like that.

P.S.: I added DE labels.

Bildschirmfoto 2025-01-07 um 09 40 06

lwjameson (approved):
Very cool.

dleadbetter:
@rsimon - Good catch on the formatting on longer content. I've updated both the MetadataList and MetadataModal components to better handle variable length content and also updated to format URLs.

## MetadataList
Screenshot 2025-01-07 at 3 48 54 PM

## MetadataModal
Screenshot 2025-01-07 at 3 49 00 PM

PR #352, opened 2025-01-08
Annotation View Fixes

rsimon:
## In this PR

This PR fixes issue 356: https://github.com/performant-software/vico/issues/356 (admin policy wasn't passed down in the annotation popup).

In addition, the PR removes two broken "integrity warnings" which would log a message when:
- the annotation had tags from someone else then the annotation's creator
- an annotation section had multiple contributors (e.g. different users for initial comment and tags)

Both cases can happen if an admin user edits someone else's annotation, and are actually legitimate.


lwjameson (approved):
Looks good

SteveMarvin:
This is accepted but I cannot move the issue.

PR #355, opened 2025-01-14
Fixes "Cannot read properties of undefined (id)" Error in Image Annotation UI

rsimon:
## In this PR

This PR fixes the error reported in https://github.com/performant-software/cove-recogito/issues/146

For posterity: the reason was a 'null' error (actually in two places), caused when the "Notes" list was open while the current `user` was undefined. This doesn't normally happen when the interface loads. But it _does_ happen if "Notes" are open while the Annotorious instance (re-)initializes – which happens when changing the page in a multipage document.

### DIFF
--- a/src/apps/annotation-image/ImageAnnotationDesktop.tsx
+++ b/src/apps/annotation-image/ImageAnnotationDesktop.tsx
@@ -252,9 +252,9 @@ export const ImageAnnotationDesktop = (props: ImageAnnotationProps) => {
     setCurrentImage(source);
   }
 
-  const onError = (error) => setToast({
+  const onError = (error: string) => setToast({
     title: t['Something went wrong'],
-    description: error.message,
+    description: error,
     type: 'error',
   });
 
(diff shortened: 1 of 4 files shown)


src/apps/annotation-image/ImageAnnotationDesktop.tsx
```diff
@@ -252,9 +252,9 @@ export const ImageAnnotationDesktop = (props: ImageAnnotationProps) => {
     setCurrentImage(source);
   }
 
-  const onError = (error) => setToast({
+  const onError = (error: string) => setToast({
```
rsimon, line 255: This showed as a type error in my IDE. (The IDE also seemed to be right about it, because the function is used from callback on the sidebar which returns a string rather than an Error object.)

src/apps/annotation-text/TextAnnotationDesktop.tsx
```diff
@@ -336,7 +336,7 @@ export const TextAnnotationDesktop = (props: TextAnnotationProps) => {
 
       <Toast
         content={toast}
-        onOpenChange={(open) => !open && setToast(null)}
+        onOpenChange={(open) => !open && setToast(undefined)}
```
rsimon, line 339: Likewise, showed up as a type error in my IDE (due to the way the toast arg is defined).

dleadbetter (approved):

lwjameson (approved):

PR #358, opened 2025-02-13
Fix for https://github.com/performant-software/cove-recogito/issues/159

lwjameson:
## In this PR

- Addresses https://github.com/performant-software/cove-recogito/issues/159

Not sure if this is the best fix, but it does work.

### DIFF
--- a/src/themes/default/annotation/toolbar.css
+++ b/src/themes/default/annotation/toolbar.css
@@ -23,6 +23,11 @@
   justify-content: flex-end;
 }
 
+.anno-toolbar-presence {
+  max-width: 12vw;
+  overflow: hidden;
+}
+
 .anno-toolbar button {
   border-radius: 50%;
   color: var(--font-dark);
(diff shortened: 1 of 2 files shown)


rsimon (approved):
I guess that will work as a (hacky) band-aid for now, at least on common resolutions and screen sized. But for a full solution, we'd need to revisit the [useCollapsibleToolbar](https://github.com/recogito/recogito-client/blob/develop/src/components/AnnotationDesktop/useCollapsibleToolbar.ts) hook.

Here's what I think is happening: the collapse threshold (at which toolbar elements drop down into the auxiliary dropdown toolbar) is only evaluated when the window resizes. We forgot to handle the case where the window doesn't resize - but contents of the toolbar change! Therefore, when users join the session, the growing presence list starts pushing the right-side buttons off the screen, without the toolbar collapse being evaluated and triggered. That's probably something we'll want to put on the backlog. 

P.S.: I also noticed that the document title doesn't display an ellipsis when truncated. Not sure what's causing this though.

src/themes/default/annotation/toolbar.css
```diff
@@ -23,6 +23,11 @@
   justify-content: flex-end;
 }
```
rsimon, line 25: Once we get a chance to build a clean solution with the `useCollapsibleToolbar` hook, we should remove this. Perhaps we can add a code comment here along those lines?

PR #362, opened 2025-02-25
Cove eula

lwjameson:
## In this PR

- This adds a feature which allows there to be a EULA that the user must accept before uploading and documents to the library.
  - If the ENV variable `PUBLIC_EULA_URL` is set, then the Accept EULA modal is displayed if the user has not yet accepted and tries to upload a document.

image

The user can choose to read the EULA and accept. If they do not accept, no upload is allowed. You only need to accept once.


src/apps/project-home/ProjectHome.tsx
```diff
@@ -18,6 +18,7 @@ import type {
   JoinRequest,
   MyProfile,
   Translations,
+  UserProfile,
```
rsimon, line 21: This one's showing as an unused import in my IDE. (I guess MyProfile is used instead.)

src/apps/project-home/ProjectHome.tsx
```diff
@@ -80,6 +83,12 @@ export const ProjectHome = (props: ProjectHomeProps) => {
     }
   }, [props.availableLayers]);
 
+  useEffect(() => {
```
rsimon, line 86: Is this ever used? The `props` shouldn't ever change, since this is the root component, and the state is already initialized with `props.user`.

src/apps/project-home/ProjectHome.tsx
lwjameson: Removed

src/apps/project-home/ProjectHome.tsx
lwjameson: Its a habit. I never trust initialized state ;-) Been caught too many times by a component being re-rendered with stale state because of some upstream change, though in this case I agree...Unlikely. Removed.

PR #364, opened 2025-03-03
Revised Plugin Framework

rsimon:
## In this PR

This PR introduces a revised plugin framework that builds on Astro Integrations.


src/backend/helpers/pluginHelpers.ts
```diff
@@ -15,15 +16,14 @@ export const getInstalledPlugins = (
...
   plugin_name: string
 ): Response<InstalledPlugin> =>
   supabase  
     .from('installed_plugins')
     .insert({
       project_id,
-      plugin_id,
-      plugin_name
+      plugin_name,
+      plugin_id: uuidv4() // Just a hack...
```
rsimon, line 26: Not sure if we need the plugin ID any more. It used to be an (arbitrary) UUID, which plugin developers were supposed to provide with their plugin. In hindsight, this feels odd from a developer perspective. 

With the new SDK, the plugin developer only declares a `name`. But perhaps we'd want a `name` and a `version`? (Both string fields.)

jamiefolsom: @rsimon including version seems like a good idea! Questions: 
- Are plugins always npm packages, and installed via insertion into `package.json`?
- Should we require them to use [semver](https://semver.org/)?
- Should the version of the plugin always be the same as the version of the NPM package (I presume)?

jamiefolsom:
Another question of possible relevance to the plugin framework. Astro 5.4.0 has some interesting [features related to configuration](https://astro.build/blog/astro-540/#new-config-helpers-and-build-options:~:text=These%20helpers%20are%20particularly%20useful%20for%20integration%20authors%20and%20for%20developers%20writing%20scripts%20that%20need%20to%20manipulate%20Astro%20configurations%20programmatically.), which might be worth looking at. Would now be the right time to update the astro version dependency @rsimon?

src/backend/helpers/pluginHelpers.ts
rsimon: <Are plugins always npm packages, and installed via insertion into package.json?>

That's right. They are always npm packages, although in production we'd probably install them via the `add` utility command that comes with Astro. (Because that will install the package + make the changes to the Astro config.)

<Should we require them to use semver?>

Makes sense.

<Should the version of the plugin always be the same as the version of the NPM package (I presume)?>

Yes. In theory we could probably read from the package.json, although it would involve a little extra hacking.

lwjameson (approved):
Looks good.

PR #365, opened 2025-03-07
String vs. Object Tags: Filter & Color Coding

rsimon:
## In this PR

When introducing the reconciliation plugin, we changed how tags are represented internally: Previously, tags were simply a string. Now they can be either a string, or an object `{ label: "a label", id: "an-id-or-uri" }`.

This PR fixes tag rendering in the __filter panel__ and the __color coding legend__, which both used to render raw JSON instead of the label.

### DIFF
--- a/src/backend/helpers/pluginHelpers.ts
+++ b/src/backend/helpers/pluginHelpers.ts
@@ -42,7 +42,7 @@ export const deleteInstalledPlugin = (
 export const updatePluginSettings = (
   supabase: SupabaseClient,
   project_id: string,
-  plugin_id: string,
+  plugin_name: string,
   plugin_settings: any
 ) =>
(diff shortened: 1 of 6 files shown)


src/backend/helpers/pluginHelpers.ts
```diff
@@ -42,7 +42,7 @@ export const deleteInstalledPlugin = (
 export const updatePluginSettings = (
   supabase: SupabaseClient,
   project_id: string,
-  plugin_id: string,
+  plugin_name: string,
```
rsimon, line 45: I just added this while I was at it. This fixes a problem that plugin settings could not be updated from the "Settings" page.

lwjameson (approved):
Looks good!

PR #368, opened 2025-04-07
W3C Annotation Export

rsimon:
## In this PR

This PR implements annotation export in [W3C Web Annotation](https://www.w3.org/TR/annotation-model/) format. Comment & reply bodies are converted from Quill to HTML.

Annotation targets are crosswalked as follows:
- Image targets are exported as either [FragmentSelector](https://www.w3.org/TR/annotation-model/#fragment-selector) (`xywh` format) or [SvgSelector](https://www.w3.org/TR/annotation-model/#svg-selector) for rectangles or polygons, respectively.
- Plaintext targets are exported as combination of a [TextPositionSelector](https://www.w3.org/TR/annotation-model/#text-position-selector) (character offsets) and a [TextQuoteSelector](https://www.w3.org/TR/annotation-model/#text-quote-selector) (annotated text snippet).
- TEI targets are exported as a combination of a [RangeSelector](https://www.w3.org/TR/annotation-model/#range-selector) with [XPathSelector](https://www.w3.org/TR/annotation-model/#xpath-selector) start and end, and a TextQuoteSelector.
- PDF targets are exported as a combination of a FragmentSelector (with `page` and `viewrect` arguments) a TextPositionSelector and a TextQuoteSelector.

Export is available for all permutations: context or project, all annotations or specific document only, with or without private annotations.


lwjameson (approved):
Looks good!

PR #370, opened 2025-04-09
IIIF Manifest Export

rsimon:
## In this PR

**Important:** this PR builds on top of PR #368, which should be merged first!

This PR adds functionality for exporting a **derivative IIIF Presentation Manifest** for an imported manifest. The derivate manifest is a copy of the original manifest, but with additional annotation pages inserted into the `Canvas` objects.

- Comment & reply bodies are converted from Quill to HTML.
- Image targets are converted to either a [FragmentSelector](https://www.w3.org/TR/annotation-model/#fragment-selector) (rectangle) or an [SvgSelector](https://www.w3.org/TR/annotation-model/#svg-selector) (polygon).

Bildschirmfoto 2025-04-09 um 14 44 17


lwjameson (approved):
Looks good.  Seems to include changes from another PR though.  Depending on the order you merge you *may* need to rebase.

Issue #372, opened 2025-04-16
PDF Export: Using tags causes errors


jsmueller7:
The PDF export function is somewhat error-prone if the content of the annotation is not expected. I have noticed that the errors occur, for example, when tags are used. (Geo annotations also cause errors; but this is an edge case that probably occurs really rarely).

rsimon:
Hi @jsmueller7, thanks for reporting. I admit I haven't actually tested PDF export in conjunction with the GeoTagger plugin yet. 

I'll take a look! As for normal tags: hm, those should work. Although they are serialized in a bit of a hacky/minimal way. (Just appended as plaintext with square brackets around them. I.e. something like: "My commment... [TagA] [TagB]" But we had some changes to the internal model recently. (E.g. the ability for tags to carry a color, or an ID/URI in addition to a label.) So perhaps that's a regression. I'll also take a look!

rsimon:
Hi @jsmueller7, looks like this happened on annotations that had a tag, but no comment. The fix will be coming within the next days! (As part of an upcoming PR which makes some enhancements & fixes to our internal tag handling.)

jsmueller7:
Hi @rsimon, yes, this happend with tags that had no comment. Thanks for fixing!

SteveMarvin:
Is this a 500 error? Still getting it...

https://recogito-staging.covecollective.org/en/projects/23fd1200-e6dc-4212-83bc-5265341d1b5c

![Image](https://github.com/user-attachments/assets/81494c2d-cbb9-4e01-8e0e-3589011970b7)

SteveMarvin:
This seems to be working fine, but I cannot move this issue. Please move it to rejected.

rsimon:
Oops, sorry. Moved it to Q&A too early. This is still an open PR.

Issue #335, opened 2024-12-09
Android issue: Annotation pop up not showing up


jsmueller7:
While trying to debug issue #334 I also tested on the staging server (netlify.app). Based upon that I'd expect that #334 persists. Additionaly the annotation pop up didn't appear any more after I had selected text. The pop up did appear though, when I seletect text through a pc and a `inspect` tab on my android device.

The header bar #281 issue fix lead to an unability to open the side bar. I have been testing with a very small screen though.

jsmueller7:
It could be. I've only tested with PDF.

jsmueller7:
I've just done some more testing. With plain text this also happens, TEI seems not to be affected.
Weird: Also #334 isn't occuring with TEI, so there might be a connection.

jsmueller7:
OK, that was too hasty. The pop-up that is not displayed relates to PDF and TXT, not TEI. But #334 only applies to PDF.

rsimon:
Sorry for the delay! This should get resolved [with this pending PR](https://github.com/recogito/recogito-client/pull/373).

Issue #334, opened 2024-12-09
Android issue 403 Error


jsmueller7:
There's an issue on android devices; I've only tested with a couple of Cromium based browsers. On the older client version `1.1.2`, if you select text and try to persist the annotation you'll get an error 403:

```
{
    "code": "42501",
    "details": null,
    "hint": null,
    "message": "new row violates row-level security policy for table \"bodies\""
}
```
If you test this in the exact same context/document with the exact same user, from a PC, you'll get a 201. This behavior is even shown if you try to usb debug from a computer. When creating the annotation with a computer mouse on the pc - **in the inspected tab on the android device** - the annotation will be created. If you try to create the annotation on the tablet / smartphone / screen, you'll get an 403.

The post-URL is the same, the authorization headers are the same.

rsimon:
Hi @jsmueller7, ok that's weird. It's definitely a permissions issue, meaning that the user's credentials are not getting through. Did you try clearing browser cache and cookies/localstorage on the Android device, to make absolutely sure there's not something intervening from our auth flow change that happened around v1.2?

Otherwise: if the credentials are leaving the device correctly, as you've confirmed with the USB debug, is there something that might happen in your Kong setup when the request comes in?

jsmueller7:
Hi @rsimon: I first tested on the older version with a device that never used the new auth flow. So that shouldn't have any influence. I can check what I see in kong. I guess though that this behavoir is reproducable on the performant staging environment as well. The other issue #335 is in the way though.

rsimon:
The timing (and, in fact, which events are thrown and which aren't) is definitely different on all platforms. (I'm tempted to say that the way different browsers behave is a mess... We've recently [revised the whole selection handling code](https://github.com/recogito/text-annotator-js/pull/151) to account for it.) 

So, I wouldn't be surprised about the timing differences. They should be ok in principle (although I suspect they play a role in the missing popup on Android and PDF).

The permissions issue reported above must be something different. 
- First, I cannot see a way how the timing would influence the permission, which would be determined by the user/auth info in the request. 
- Second: I'm now realizing the message is from the `bodies` table. When you create a new annotation, this won't create a body just yet. What happens is that records are created in the `annotations` table, and the `targets` table. But an insert in the `bodies` table would only happen when you type a test message and hit Save.

jsmueller7:
With PDFs in the mobile browser I have this flow:

* First I select text and the context menu appears (copy, web search etc.)
[...]

PR #373, opened 2025-04-21
Enhancements: Core Support for Semantic Tags & Bugfixes

rsimon:
## In this PR

This PR makes our existing, half-finished support for semantic tags more solid.

- Tags in an annotation are now generally handled via the **VocabularyTerm** type, a JS object which:
   -  always has a `label`.
   -  may have a string `id` (could be a URI, but doesn't necessarily have to be).
   -  may have a pre-assigned `color`.
- The **Project Settings** page was updated, so that the user can now (optionally) provide IDs for terms when adding to the  vocabulary.
- Additional functionality was added to the **TEI crosswalk** to properly handle semantic tags in a standards-compliant way.
- Additional functionality was added to the **PDF crosswalk** to surface the tag ID in PDF viewers.
- For **backwards compatibility**, all functions that handle tags (display/autosuggest/filter/color-code/color-legend in the annotation view, export crosswalks) will **still** handle cases where a term/tag is a plaintext string. This way, old annotations already in the database from older Recogito versions won't break the system.

### Additional Enhancements and Bugfixes:
- Issue #372 was fixed as part of revising the PDF crosswalk.
- Minor versions were upgraded on some package dependencies.
- Specifically, upgrades of the `text-annotator`/`pdf-annotator` packages brings some bugfixes related to Android support, fixing issues #334 and #335.


src/apps/annotation-image/LeftDrawer/LeftDrawer.css
```diff
@@ -14,6 +14,10 @@
   height: 100%;
 }
 
+.ta-left-drawer aside .tablist {
```
rsimon, line 17: Annotation view: I noticed that when the left sidebar was scrollable, the sidebar contents would show underneath the tablist. Made that fix while I was at it.

src/backend/helpers/tagHelpers.ts
```diff
@@ -162,6 +171,7 @@ export const getTagDefinitions = (
       metadata,
       tags (
         id,
+        created_at,
```
rsimon, line 174: Note: this was missing and caused a TS warning.

src/util/export/tei/teiExporter.ts
```diff
@@ -77,32 +78,45 @@ const createTaxonomy = (annotations: SupabaseAnnotation[], document: any) => {
...
+    } catch {
+      return str;
+    }
+  });
 
   // Note: we'll later want to use URIs for tag IDs. Slugifying
   // the label is going to be a fallback option
-  const idFromLabel = (label: string) =>
-    slugify(label, {
+  const idFromTerm = (term: string | VocabularyTerm, taxonomyId: string) => {
```
rsimon, line 101: For the TEI taxonomy element, we'll use the ID of the tag if any, or a slugified version of the label.

src/components/AnnotationDesktop/AnnotationPopup/AnnotationPopup.tsx
```diff
@@ -72,7 +73,7 @@ export const AnnotationPopup = (props: AnnotationPopupProps) => {
       key={selected.id}
       className="annotation-popup not-annotatable">
       <AnnotationCard 
-        autoFocus
+        autoFocus={!isMobile()}
```
rsimon, line 76: This keeps the text selection editable on mobile touch devices.

lwjameson (approved):
Looks good!

PR #377, opened 2025-05-19
Plugin Extension Point Changes

rsimon:
## In this PR

This PR makes a few changes and additions to the plugin extension points, so we can cover all the requirements of the new plugins:

- New extension point to inject Annotorious plugins into the (image- or text-) annotator. 
- For extensions on the annotation (popup or sidebar) card, the extension point is no longer on the annotation card __section__, but the card itself.

The PR also includes some minor JS and CSS fixes.


package.json
```diff
@@ -15,50 +15,50 @@
...
     "@radix-ui/themes": "^3.2.1",
-    "@react-spring/web": "^9.7.5",
-    "@recogito/annotorious-supabase": "^1.3.2",
-    "@recogito/pdf-annotator": "^1.0.2",
-    "@recogito/react-pdf-annotator": "^1.0.2",
-    "@recogito/react-text-annotator": "^3.0.2",
-    "@recogito/studio-sdk": "^0.5.1",
-    "@recogito/text-annotator": "^3.0.2",
-    "@recogito/text-annotator-tei": "^3.0.2",
+    "@react-spring/web": "^10.0.0",
```
rsimon, line 45: This was the last dependency that kept us from switching to React 19. I updated this, but kept React at v18. React 19 is something we should test more thoroughly first...

src/components/AnnotationDesktop/SelectionURLState/SelectionURLState.tsx
```diff
@@ -32,6 +32,8 @@ export const SelectionURLState = (props: SelectionURLStateProps) => {
 
     if (selected.length > 0) {
       hash.set('selected', selected.join(','))
+    } else {
+      hash.delete('selected');
```
rsimon, line 36: Small issue I noticed: deselecting an annotation in the interface did not clear the URI hash.

PR #378, opened 2025-05-21
Plugin Framework and Extension Points

rsimon:
## In this PR

Applies various tweaks to the extension point interfaces.
- Dropped the (unused) `annotation:*:tag-autosuggest` extension point.
- Instead, there's now new `annotation:*:taglist` extension point: 
   - This extension point has the same interface as `annotation:*:annotation-editor'`.
   - It is placed alongside the default tag editor.
   - It only shows when the default tag editor shows, i.e. only on the first card section, when the annotation is empty or in "editing" mode. (The Getty plugin now uses this extension point!) 
- Added the missing `isSelected` prop on the `EmptyAnnotation` extension mount. (The GeoTagger button did not appear on a new annotation because of this!)


lwjameson (approved):
Looks good.

PR #381, opened 2025-05-30
COVE: linking between annotations in the same page

rsimon:
## In this PR

This PR improves support for links to annotations inside comments. Previously, a link to an annotation would **always** open a new tab. With this PR, link behavior will differ depending on whether the link is to an annotation on the same page or not:
- If on the same page, the linked annotation will be selected.
- In text mode, the interface scrolls to the annotation.
- In image mode, the interface zooms to the annotation.

**Limitation:** clicking a link to an annotation in the same IIIF document, but on a different **page**, will still open a new tab. (This is, unfortunately, a trickier problem to solve.)

Note: this PR also fixes a bug in the IIIF image view which caused the interface to break when an initial selection pointed to an annotation that was not on the first page of the IIIF.


lwjameson (approved):
Looks great!

PR #383, opened 2025-06-09
Assignment document order

lwjameson:
# Summary

Adds the ability to set the order of documents in assignments via drag-and-drop.  

- Addresses https://github.com/performant-software/cove-recogito/issues/169


rsimon (approved):
Looks good!

PR #384, opened 2025-06-10
PDF Export: Additional Error Handling

rsimon:
## In this PR

This PR adds an extra safeguard against broken PDF annotations.

The [issue reported here](https://github.com/performant-software/cove-recogito/issues/176#event-18061051493) was caused by an annotation without a text selection (causing the `quadpoints` field in the annotation to be empty). The annotation is also visible in the annotation view sidebar. We need to investigate **how and why** such annotations get created in the first place. This PR simply prevents the export from crashing if such an annotation is in the document.


lwjameson (approved):
Looks good!

PR #389, opened 2025-06-30
Self-Hosting Regression: OSD 5 Auth Token for info.json

rsimon:
## In this PR

The [upgrade from OpenSeadragon 4 to 5](https://github.com/recogito/recogito-client/commit/64778320ca3a75a2ddc195839786d7e3ac8f2c46) introduced a regression for self-hosted setups: self-hosted setups require that OSD sends the auth token in the `Authorization` header along with requests to `info.json` as well as image tiles. However, this no longer worked in OSD 5.

This PR changes the OSD config to the pattern recommended in [this OSD issue](https://github.com/openseadragon/openseadragon/issues/2659).


lwjameson (approved):
Looks good.

PR #392, opened 2025-08-12
TEI Export: "Documents" vs. "Assignments" Tab Behavior

rsimon:
## In this PR

This PR fixes the export behavior for all export formats (TEI, PDF, CSV, W3C, IIIF).

- In the **Documents** tab, all annotations, **from all layers on the document** are exported.
- In the **Assignments** tab, only the annotations **from all layers on the document in this assignment** are exported.

Previously, behavior was broken when exporting the base layer assignment: this export always included all layers, regardless whether exported from the **Documents** or **Assignments** tab.

### DIFF
--- a/src/apps/project-home/AssignmentsView/AssignmentDetail/AssignmentDetail.tsx
+++ b/src/apps/project-home/AssignmentsView/AssignmentDetail/AssignmentDetail.tsx
@@ -7,12 +7,10 @@ import type {
   UserProfile,
   Document,
   Group,
-  ContextDocument,
 } from 'src/Types';
 import {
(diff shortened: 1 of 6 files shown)


src/apps/project-home/AssignmentsView/AssignmentDetail/AssignmentDetail.tsx
```diff
@@ -7,12 +7,10 @@ import type {
   UserProfile,
   Document,
   Group,
```
rsimon, line 9: Removed a few unused imports I stumbled upon.

PR #393, opened 2025-08-13
Fixed broken export TEI base layer export

rsimon:
## In this PR

This PR fixes a bug that would case TEI base layer assignment exports to break if the TEI filename did not end with `.xml`...


src/util/export/sanitizeFilename.ts
```diff
@@ -3,7 +3,7 @@ import slugify from 'slugify';
 
 /** Make sure we don't have unsafe or unicode characters in download filenames **/
 export const sanitizeFilename = (unsafe: string) => {
-  const slugified = slugify(unsafe, {
+  const slugified = slugify(unsafe || '', {
```
rsimon, line 6: Being extra defensive here...

PR #396, opened 2025-08-29
Assy fixes

lwjameson:
# Summary
- Fox being able to close the nag dialog through the keyboard
- First try at getting document cards to have `aria` labels
  - Note, I am not sure if we currently do this in the image annotation view or not. The note was ambiguous.


rsimon (approved):
Looks good!

PR #402, opened 2025-09-18
Responsive Toolbar Fix

rsimon:
## In this PR

This PR addresses issue [198](https://github.com/performant-software/cove-recogito/issues/198).

- The toolbar collapse rule (i.e. whether toolbar elements should be moved into the secondary dropdown menu) is evaluated immediately after page load, not on first resize.
- Max. number of presence avatars is now 4. (All others will go into the dropdown menu, as before).
- Minor fixes to CSS rules for correct flexbox behavior.


src/apps/annotation-text/AnnotatedText/AnnotatedText.tsx
```diff
@@ -171,6 +171,7 @@ export const AnnotatedText = (props: AnnotatedTextProps) => {
 
           {props.usePopup && (
             <TextAnnotationPopup
+              asPortal
```
rsimon, line 174: Note: this will ensure identical popup move/scroll behavior with the latest versions of `@recogito/text-annotator`.

src/components/AnnotationDesktop/useCollapsibleToolbar.ts
```diff
@@ -22,15 +22,16 @@ export const useCollapsibleToolbar = () => {
...
+
+    setTimeout(() => {
+      const shouldCollapse = ref.current ? ref.current.scrollWidth > ref.current.clientWidth : false;
+      if (shouldCollapse && !collapsed) {
+        setBreakpoint(windowWidth);
+        setCollapsed(true);
+      } else if (!shouldCollapse && windowWidth > breakpoint) {
+        setCollapsed(false);
+      }
+    }, 1);
```
rsimon, line 34: Evaluates collapse condition after rendering is complete.

Issue #404, opened 2025-09-26
Various dependency updates


rsimon:
This is a todo list on the various dependencies that need updating. I already [created a PR](https://github.com/recogito/recogito-client/pull/403) that makes a number of minor/patch release updates which all looked safe.

So far, I didn't touch the following dependencies below, since these will need a bit more care & testing:

- [x] Astro had several minor version updates (5.7 → 5.14). I don't think there are breaking changes though. (@blms)
- [x] `@supabase/ssr` had a minor version update (0.6 → 0.7). (@blms)
- [x] `@supabase/supabase-js` had multiple minor version updates (2.49 → 2.58). (@blms)
- [x] `@uppy/core` and `@uppy/xhr-upload` had major version upgrades (4 → 5). **Note:** we previously had a problem with a regression bug in uppy/xhr-upload, which broke image uploads to IIIF cloud, and had it pinned to v3.6.1 for a while. This bug seems to have been resolved in v4 (which we are currently using), but it's definitely something we should test for. (@blms)
- [x] `nodemailer` had a major version upgrade (6 → 7) (@blms)
- [x] `react-window` had a major version upgrade (1 → 2). We are using this in the IIIF thumbnail view and I think there are breaking API changes (@rsimon).
- [x] `es-lint`(dev dependency) had minor version updates, which are probably safe. (@blms)
- [x] `typescript` (dev dependency) had a minor version update, which is probably safe (@rsimon).

**Upgrades on our own libraries:**
- [x] `@recogito/text-annotator`, `@recogito/text-annotator-tei` and `@recogito/react-text-annotator`. These had a minor version update with changes to selection handling and the React text popup, and improvements to the highlight renderer . There are no breaking changes, but we should thoroughly test for regressions in plaintext and TEI views. The text popup also has a new `asPortal` prop which we (probably) need to enable to maintain current behavior (@rsimon).
- [x] `@recogito/pdf-annotator` and `@recogito/react-pdf-annotator`. There are no breaking changes either, but we should test for regressions – especially because the new version includes a major PDF.js update (PDF.js 4 → 5). Also: the new renderer has additional init options for highlight smoothing we should leverage (@rsimon).
- [x] `@recogito/studio-sdk`. Shouldn't have any breaking changes. But we should test with the existing plugins (@rsimon).
- [x] `cozy-iiif` had minor version updates (0.2 → 0.4). I think all of them are safe, but will double-check (@rsimon).

blms:
Closed by release 1.8.0

PR #408, opened 2025-09-30
Dependency updates: Astro, supabase, Uppy, nodemailer

blms:
### In this PR

Per #404:
- Update Astro to latest 5.14.1
- Update `@supabase/ssr` to latest 0.7.0 and `@supabase/supabase-js` to latest 2.58.0
- Update `@uppy/core` to latest 5.0.2`@uppy/xhr-upload` to latest 5.0.1
- Update `nodemailer` to latest 7.0.6

Other:
- Addresses a race condition that sometimes happens in uploading; it seemed to happen more consistently after the Uppy upgrade but I noticed it before, hard to tell if it's related or not.
   - Basically Uppy would sometimes fire a `progress` event triggering the `onProgress` callback at 100% progress,  _after_ the `onSuccess` callback, setting the upload in the UI back to an in-progress state. 
   - Since we only really need a visual progress update at less than 100%, and then expect the success callback once it reaches 100%, I figured it was fine to just not mutate state if the callback runs when the progress percent is 100.
- Fixes a bug where long title names could overflow the upload tracker

### Notes
- No applicable breaking changes in Astro, Uppy, or supabase deps
- I don't have email set up locally to test `nodemailer`, but the [only breaking change](https://github.com/nodemailer/nodemailer/releases/tag/v7.0.0) appears to pertain to a [feature we don't use](https://nodemailer.com/transports/ses). But we can test it in QA.


rsimon:
Looks good – and good catch about the upload glitch!

Issue #409, opened 2025-09-30
Allow Org Admin users to manage collections within an instance


blms:
### testing notes

On the [Recogito staging site](https://recogito-staging.covecollective.org/):
- [ ] Log in as an Org Admin
- [ ] Click on your user icon in the top right. There should now be an option in the dropdown to manage collections
- [ ] Clicking on it should bring you to a screen where you can create a new collection
- [ ] After creating, you should be able to edit the collection's name or delete it
- [ ] Search should work to filter the list of collections
- [ ] Click on the collection and you should be able to import documents to it
- [ ] Import some documents and they should appear in the list
- [ ] On a row for a single document, you should be able to import revisions (i.e. an updated version of the same text)
- [ ] You should also be able to edit their metadata or delete them (revisions are deleted one at a time)
- [ ] It should be possible to filter this list with the search
- [ ] When you return to the collections page, it should correctly count the number of documents within each collection
- [ ] Return to the main projects list page and in one of your projects, go to add a document. Your new collection(s) should appear in the left sidebar of the document library
- [ ] You should be able to add documents from the new collection
- [ ] Once you've added a document from a collection to your project, you should not be able to delete that collection. It should mention in the error that perhaps the document has been used in a project
- [ ] The same thing should happen if you try deleting the document itself

SteveMarvin:
This seems to be another issue that I can't move. Can someone please shift this to ACCEPTED?

PR #410, opened 2025-09-30
Basic collection management (#409)

blms:
### In this PR

Per #409:
- Collection management by org admins, available from the user profile dropdown
- List of collections, with document counts
- Click on a collection to list all documents in it, and import new documents into it
   - From here, edit document metadata
   - Upload new revisions of documents
   - Archive documents

### Notes

This requires migrating the `recogito-server` supabase instance to the version [on this branch](https://github.com/recogito/recogito-server/pull/60).

### Questions

1. What actions should be available on project documents? So far I have editing title/metadata, uploading revisions, and deletion.
   - Should it be possible to view a document here, outside of the context of a project/assignment, and if so would we need to prevent annotation?
   - How about listing/viewing revisions? This is possible in the document library already, but wondering if it makes sense here
   - COVE documents come with Author/Artist info. Do we want to let users add that here?
2. Is there going to be any problem with mixing content types (IIIF, txt, img) in a single collection, and/or in a revision of a document?
    - I guess we also need to enforce that only 1 file can be uploaded for a revision? Edit: Just added this in 216992f.
3. Do we need EULA stuff in the `CollecitonManagementDocumentActions` form for adding a revision? I guess at that point they would already have to have accepted the EULA to upload the original document, but theoretically it's possible they haven't.
4. How should we generate `collection_metadata.document_id`—does the method I've chosen make sense? It probably ought to be different from the document's UUID, but we also want it to be unique since it's used to create revisions.
5. How should we handle deleting documents with multiple revisions? My initial approach is to allow an org admin to delete the latest revision one at a time, since it's possible that a prior revision is in use in a project.


blms:
Ok, this one is ready for review with lots of open questions 😅 

I think for the most part they are minor questions and the basic functionality is fully in place.

rsimon (approved):
This looks great!

Just two questions/comments:
- Do we want the option to delete a collection (or "archive" - if that's possible in the backend?)
- I think I'd remove the "Create Collection" button from the view that shows the documents for a specific collection. Logically, creating a new collection makes the most sense while the user is looking at the collections table (rather than the table of documents in another collection).
- Do we want the option to import **existing** documents in to a collection? Currently, upload is the only option to include a doc into a collection.

PR #411, opened 2025-10-01
Upgrade ESLint and resolve linter errors (#404)

blms:
### In this PR

Per #404:
- Upgraded ESLint to latest 9.36.0
   - Used the automated tool (`npx @eslint/migrate-config`) to migrate our config into the new flat format, and installed the required dev dependencies
   - Manually converted the migrated config from CJS into ESM, which is the recommended module type
- Ran linter on the command line and applied both automatic and manual fixes to errors
- Removed abandoned `@types/react-window` as types are now included in the original package

### Notes / Questions

- Should we have some kind of pre-commit hook to lint any changes before they are committed? (maybe code style too?)
- In `AssignmentDetail`, the `activeId` state value is never actually used, though it's set in a few places. Should we remove everywhere that it's set, or is it intentional to set it without using its value anywhere?
- In `AnnotatedTEI/behaviors.ts`, ESLint complained about non-null assertion after optional chaining. I figure the point is that this will never be null, but can't hurt to have a fallback to an empty string just to make both ESLint and TS happy.
   - It also complained about this in `FilterPanel/FilterState.tsx`. 
- I removed the`ColorPalettes.shuffle` function, which ESLint complained about due to the use of `T extends any` but appears to be unused.


rsimon:
CC-ing @lwjameson on the `activeId` question. I think this may simply be a leftover from a previous version of the AssignmentDetail component?

> In `AssignmentDetail`, the `activeId` state value is never actually used, though it's set in a few places. Should we remove everywhere that it's set, or is it intentional to set it without using its value anywhere?

rsimon (approved):
A big one! Thanks for this - all looks good!

blms:
On further investigation, it looks like the `activeId` state was never used, maybe just some cruft from WIP work. Drag and drop functionality works the same with or without `setActiveId` as far as I can tell, which makes sense since setting an unused state doesn't have much effect.

Made that change along with some minor CSS tweaks for the assignment creation table.

rsimon (approved):

rsimon:
> do you have any thoughts about enforcement of linting/formatting?

TBH I don't have a strong opinion. But a linting pre-commit hook might be a good thing. (Also CC-ing @lwjameson!)

About the formatting: we actually have a `.prettierrc` file that's perhaps worth revisiting. The style in the current codebase isn't perfectly uniform, I think, but nothing that would worry me.

blms:
@rsimon Thanks, don't know how I missed the `.prettierrc` that's already there! 🤦 

I also have no strong opinion and am eager to get the dependency upgrades out at this point, so I'm probably going to go ahead and merge this unless @lwjameson objects

PR #419, opened 2025-10-20
Allow importing existing documents from a project into a collection (#418)

blms:
### In this PR

Per #418:
- New option in the dropdown for importing in the collection management UI that allows importing from an existing project
   - New read-only prop for the `DocumentLibrary` to enable this, which prevents documents from being modified/imported there; also a prop to hide the Collections panel
- Calls to the new backend RPC to actually make the copy/copies and set them on state

### Notes

- Requires the latest migration from https://github.com/recogito/recogito-server/pull/65 to work.
- Let me know if the German translations need any work!

### DIFF
--- a/src/apps/collection-management/Collection/Collection.tsx
+++ b/src/apps/collection-management/Collection/Collection.tsx
@@ -1,16 +1,20 @@
 import { useEffect, useMemo, useState } from 'react';
 import { v4 as uuidv4 } from 'uuid';
-import { archiveDocument, updateCollection } from '@backend/crud';
+import { archiveDocument, getDocument, updateCollection } from '@backend/crud';
 import { supabase } from '@backend/supabaseBrowserClient';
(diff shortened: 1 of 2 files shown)


rsimon (approved):
Looks good!

blms:
@rsimon Thanks!

Ok, in combination with the additional new migration from https://github.com/recogito/recogito-server/pull/65, this should now be in better shape.

I did as you suggested by pulling file contents before actually attempting the copy, and archiving any documents that failed to upload file contents.

I also realized I needed slightly different handling for remote IIIF, so I made sure that was handled as well. (Basically it just meant skipping the file download/upload steps, and appears to work for me locally.)

blms:
Also this code is looking a little unwieldy… let me know if you have any quick ideas about how to pull some of this out of the single function.

blms:
Probably fine as is for now in that case. Thanks!

PR #420, opened 2025-10-21
Allow usersnap focus while inside dialogs

blms:
### In this PR

Per performant-software/cove-recogito#203:
- Prevent Radix dialog focus trap (which is an accessibility feature) when Usersnap is being focused


rsimon (approved):
Looks great!

PR #421, opened 2025-10-24
Switch email logo to png

blms:
### In this PR

- Switch the email templates' SVG logo to a PNG, as most webmail clients (including Gmail) do not support inline SVGs.

### Questions

Do you happen to know anything about the integration of these email templates? It looks like the work to create them was done in #297, but it's unclear to me exactly how they were meant to be used.

My general impression is that they should just be copy-pasted directly into the Supabase auth email templates section. COVE Studio ran into errors because they were still on an old template, and copy-pasting this reset password template (I guess using the new auth flow) into their instance seems to have fixed it.

If that's the case, I'll also need to update the docs for [configuring the email templates](https://github.com/performant-software/recogito-cloud/blob/develop/create-tenant.md#configure-email-template).


blms:
Actually I noticed that @camdendotlol made the change to the deploy instructions here https://github.com/performant-software/recogito-cloud/commit/bd1b5a3cc8a124d936a43efcac859b99fcf6d20c. 

Camden would you happen to know any more info about this template or how it was meant to be used? The url as written in the recogito-cloud docs: 
```
{{ .ConfirmationURL }}/reset-password
```
was causing errors, while the URL from the `template-reset.html` template in this repo
```
{{ .SiteURL }}/auth/confirm?token_hash={{ .TokenHash }}&type=recovery&next=/en/reset-password
```
seems to be working.

camdendotlol:
I don't really know anything about the template email or how it's used. In that commit I was rearranging sections a bit, but I left that line intact. If the new version works, it's probably worth updating the readme to use it instead.

PR #430, opened 2025-11-06
PDF Annotator React 18/19 compatibility fix

rsimon:
## In this PR

As discussed via Slack, I cut a new release for the PDF Annotator that includes [the React 18/19 dual compatibility fix](https://github.com/recogito/pdf-annotator-js/pull/2). In local testing, this fix now works well with React 19 for all text annotation interfaces (plaintext, TEI, PDF).


blms (approved):
Thanks for getting the quick release out!

Issue #431, opened 2025-11-14
Migration to i18next


blms:
Tracking work to migration from custom JSON solution to i18next.

PR #432, opened 2025-11-14
Migration from custom JSON to i18next (#431)

blms:
### In this PR

Per #431: Migrate all existing translations to i18next.
- Conversion of all `props.i18n.t['key']` to `t('key', { ns: 'namespace' })` (accomplished with a small python script), which eliminates the need to pass `i18n` as props to translate strings
   - Update `t[variable]` calls to use the correct namespace, or remove them when translations may not exist (e.g. errors directly from postgREST)
- Conversion of all `props.i18n.lang` to use i18next's built-in `i18n.language`, which eliminates the need to pass `i18n` as props to get the current language
- Move translation files into `public/` so that we can use the HTTP backend for react components
- Use [namespaces](https://www.i18next.com/principles/namespaces) in i18next so that we can maintain separate JSON files for translations, and not load all of them at once
   - Load them individually per-component using `useTranslation([ns1, ns2])`
- Wrap all root-level React components in `I18nextProvider` so they can use the `useTranslation` hook (these are `react-i18next` things)
- Use the `getFixedT` solution from [here](https://www.divby0.io/posts/using-i18next-in-astro/) in server-rendered Astro components and Astro APIRoutes, using the filesystem backend
- Detect the language from the `/[lang]/` URL path segment; at the root URL, detect based on user's browser preferred language
- Update all string interpolation to use i18next's built-in interpolation

### Notes
- One unfortunate drawback to i18next is that as far as I can tell, when we're using multiple namespaces (i.e. multiple JSON files), we have to specify per-translation which namespace we're pulling from.
   - I always use the `t('key', { ns: 'namespace' })` style as preferred by i18next. It is also possible to use `t('namespace:key')` but that seemed clunkier for the sentence-length keys. There may also be some fallback behavior where we don't always need to specify the namespace, but I didn't look into it in the interest of saving time. (Probably useful to check when developing new features though.)
- When there were keys duplicated across multiple JSON modules, I had started by picking which one seemed the most appropriate. But once I realized there were a few hundred cases of this in code, I ended up just putting them into `common.json` and preferring that as the namespace. It was easier to accomplish the goal with scripts that way. At some point these can probably be removed from the other namespaces, but we would need to make sure all references to them were updated in code.
- I also filled in a few missing keys and deleted a few unused ones. Not exhaustively, though.
- Generally, i18next leads us to prefer camelCase for naming keys, as underscores are used for things like combining pluralized keys (see `selectedDocuments_one`, `selectedDocuments_other`). I didn't change this everywhere though, just where it was most applicable for plurals, and a few other areas.

### Questions

- Am I missing anything for language detection and switching? As far as I could tell, browser language was only checked at the root URL and then all internal links use the current language; and a user could only switch languages by changing the URL manually. I just replicated this, but if I'm mistaken it shouldn't be hard to adjust.

### DIFF
--- a/src/apps/accept-invite/AcceptInvite.tsx
+++ b/src/apps/accept-invite/AcceptInvite.tsx
@@ -64,7 +64,7 @@ const AcceptInvite = (props: AcceptInviteProps) => {
   );
 };
 
-export const AcceptInviteWrapper = (props: AcceptInviteProps) => (
+export const AcceptInviteApp = (props: AcceptInviteProps) => (
   <I18nextProvider i18n={clientI18next}>
     <AcceptInvite {...props} />
   </I18nextProvider>
--- a/src/apps/accept-org-invite/AcceptOrgInvite.tsx
+++ b/src/apps/accept-org-invite/AcceptOrgInvite.tsx
@@ -13,7 +13,7 @@ interface AcceptOrgInviteProps {
   token: string;
 }
 
-const AcceptOrgInviteComponent = (props: AcceptOrgInviteProps) => {
+const AcceptOrgInvite = (props: AcceptOrgInviteProps) => {
   const { t, i18n } = useTranslation(['auth-reset-password']);
 
   const [password, setPassword] = useState('');
@@ -138,8 +138,8 @@ const AcceptOrgInviteComponent = (props: AcceptOrgInviteProps) => {
 };
 
 
-export const AcceptOrgInviteWrapper = (props: AcceptOrgInviteProps) => (
+export const AcceptOrgInviteApp = (props: AcceptOrgInviteProps) => (
   <I18nextProvider i18n={clientI18next}>
-    <AcceptOrgInviteComponent {...props} />
+    <AcceptOrgInvite {...props} />
   </I18nextProvider>
 );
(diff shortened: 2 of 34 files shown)


rsimon (approved):
This is fantastic! A huge effort, and such an improvement over our previous i18n hack. Plus, getting rid of all those drilled `i18n` props really lightens the code I think. Thanks for this!

One thing I noticed: while in DE language mode, the client-rendered pages fetch both the DE and the EN translations. But I guess that's the correct behavior, since the client will always fetch the current locale (DE), plus the default locale (EN) for fallbacks, right?

The other minor nitpicky thing (really mostly for the sake of providing at least _one_ piece of actual feedback ;-): we now have lots of `...Wrapper`  components, because of the new i18react context. I wonder if I would have reversed the naming convention, i.e. using the original name for the root component, and a different name for the inner one. I.e. something like:

```
<CollectionManagement>
  <LocalizedCollectionManagement />
</CollectionManagement>
```

or perhaps

```
<CollectionManagementApp>
  <CollectionManagement />
</CollectionManagementApp>
```

...just because it felt a bit odd to import a `...Wrapper` in the root page astro file. But–as I said–really not a true issue, as long as we're using a common convention across the codebase.

rsimon:
Looks great!

PR #435, opened 2025-11-20
#375 - Data export/import

dleadbetter:
This pull request implements the ability to export project data to a zip file containing CSVs and binary data. This pull request adds the new `jobs` table which will serve as a placeholder for the zip data to be downloaded. Jobs will run in the background via Trigger.dev. Once a job is completed the resulting zip file can be downloaded via the Jobs list.

### Export Project
Screenshot 2025-11-20 at 11 26 17 AM

### Jobs Navigation
Screenshot 2025-11-20 at 11 26 28 AM

### Jobs List
Screenshot 2025-11-20 at 11 38 01 AM

This pull request also implements the ability to import a zip file containing JSON and binary data which will create a project and all other necessary data (annotations, users, layers, etc). On import, new IDs will be created for all data except `documents` and `profiles`. Documents will be matched by ID, profiles will be matched/updated by email address. All new users will be required to change their password and login via local authorization.

### Import Project
Screenshot 2025-12-09 at 3 05 07 PM

### Jobs List
Screenshot 2025-12-09 at 3 05 19 PM

### DIFF
--- a/src/trigger/importProject/index.ts
+++ b/src/trigger/importProject/index.ts
@@ -11,7 +11,7 @@ interface Payload {
   token: string;
 }
 
-const IIIF_KEY = process.env.IIIF_KEY || process.env.RECOGITO_TIGER;
+const IIIF_KEY = process.env.IIIF_KEY;
 const IIIF_URL = process.env.IIIF_URL;
 const IIIF_PROJECT_ID = process.env.IIIF_PROJECT_ID;
 
--- a/src/pages/api/images.ts
+++ b/src/pages/api/images.ts
@@ -2,7 +2,7 @@ import { getMyProfile } from '@backend/crud';
 import { createSupabaseServerClient } from '@backend/supabaseServerClient';
 import type { APIRoute } from 'astro';
 
-const IIIF_KEY = import.meta.env.IIIF_KEY || import.meta.env.RECOGITO_TIGER;
+const IIIF_KEY = import.meta.env.IIIF_KEY;
 const IIIF_URL = import.meta.env.IIIF_URL;
 const IIIF_PROJECT_ID = import.meta.env.IIIF_PROJECT_ID;
 


blms (approved):
Code looks great! The approach looks solid to me. Haven't run it locally but let me know if that would be helpful.

Also, let me know if you need help with the new i18next stuff to resolve conflicts—PR #432 may also help explain some of it.

trigger.config.ts
```diff
@@ -0,0 +1,22 @@
+import { defineConfig } from '@trigger.dev/sdk/v3';
+
+export default defineConfig({
+  project: 'proj_laigugtxozpzhsnwklfr',
```
dleadbetter: Should this be configurable via environment variable? Or will each instance use the same Trigger.dev project?

blms: I would say yes, it should be configurable. Even if all of our self-managed instances use the same project, people may be self-hosting.

Does the Trigger.dev side need any configuration that we should document somewhere, or will it just work with a bare Trigger.dev project using this config?

src/trigger/importProject/index.ts
```diff
@@ -0,0 +1,308 @@
...
+import { generatePassword } from '@util/auth';
+import AdmZip from 'adm-zip';
+import { v4 as uuidv4 } from 'uuid';
+
+interface Payload {
+  jobId: string;
+  token: string;
+}
+
+const IIIF_KEY = process.env.IIIF_KEY || process.env.RECOGITO_TIGER;
```
blms: I see this `RECOGITO_TIGER` somewhere else too, any idea what it is? (if not, can probably be removed from both places)

rsimon: I have no idea what that is... the other place is in the API handler that forwards the image uploads to IIIF cloud. But I don't think I put it there. Looks like it came in with [this commit](https://github.com/recogito/recogito-client/commit/2112eb022eb73ec8d3400e9c8bcac824818cdca9).

src/backend/storage/supabase.ts
```diff
@@ -16,6 +18,8 @@ export const uploadFile = (
   name: string,
   onProgress?: (progress: number) => void
 ): Promise<void> => new Promise((resolve, reject) => {
+  // TODO: We have no session here
+  // TODO: Exract uppy code and pass token?
```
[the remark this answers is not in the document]
dleadbetter: Ah, yup, just ended up using the Supabase client. Removed.

blms (approved):

PR #440, opened 2026-01-06
Handle null and non-array selectors in pdf export

blms:
### In this PR

Fixes a bug users were encountering where `annotations.filter(annotation => (annotation.target.selector as PDFSelector[]).every` was throwing an error on `.every` when the selector was `null`.

This PR just puts in some safeguards to handle `null` selectors, as well as some other potential problems (single selector, missing/NaN page number).

I'm not sure what might have caused the selector to become null, so that may require some digging if we want to diagnose. This PR just allows the PDF export to happen regardless. FWIW, the record was this one: https://unc.covecollective.org/en/annotate/99e259df-3724-4ac3-a0ee-ecb7802a6fda/5f486654-e902-4884-b728-a9843b795603


rsimon:
Yikes. Will take a closer look tomorrow. But: a "Note" would be an annotation without a selector.

PR #457, opened 2026-02-10
Add skip to main link

camdendotlol:
# Summary

This PR adds a new "Skip to main content" link that appears when tabbing through the UI, and adds an ID value of `main` to the main content element for most/all of the pages. In many cases, the main content was already marked using the `main` element type, but when that was not the case I converted the element to `main`.

Some of the placement was a bit subjective, e.g. if the page shows a list of records, should we skip to the records themselves or to the inner header buttons? I scoped pretty aggressively in those cases.

## Screenshot

Screenshot 2026-02-10 at 1 57 38 PM

### DIFF
--- a/src/pages/500.astro
+++ b/src/pages/500.astro
@@ -6,7 +6,7 @@ const locale = Astro.currentLocale;
 const t = await getFixedT(locale, ['error']);
 ---
 <BaseLayout title="Recogito | Error">
-  <main class="container" id='main'>
+  <main class="container" id="main">
     <div class="column">
       <h1>500 Error</h1>
       <p>{t('Something went wrong.', { ns: 'error' })}</p>


src/pages/404.astro
```diff
@@ -1,20 +1,20 @@
...
-const lang = Astro.currentLocale || (preferred && supported.includes(preferred) ? preferred : 'en');
-const t = await getFixedT(lang, ['error']);
+const locale = Astro.currentLocale;
+const t = await getFixedT(locale, ['error']);
 ---
 <BaseLayout title="Recogito | Not Found">
-  <div class="container" style="align-items: center; bottom: 0; display: flex; justify-content: center; position: absolute; top: 0; width: 100%;">
+  <main
+    class="container"
+    id='main'
```
blms: One minor request: could we make single vs double quotes consistent within the same file? 

From looking at this, it seems the general convention is to use double quotes for HTML attrs in astro files, and single quotes elsewhere. (Though that may already be inconsistent if we don't have formatting rules for Astro in Prettier/ESLint.)

PR #493, opened 2026-04-25
Move annotation tags to `annotation` elements and deal with URI encoding

ajolipa:
### In this PR
- Moves the list of tags for an annotation from separate `rs` elements to the `ana` attribute on the `annotation` element itself. This is more compatible with the FairCopy software TEI handling.
- Applies URI encoding/decoding to the individual tags to handle cases where tags contain spaces or other special characters. (Screenshot from local testing.) 
image

Exported TEI from local testing:
```
<standOff type="recogito_studio_annotations">
  <listAnnotation>
    <annotation ana="#test #this%20is%20a%20test #//&??" target="/TEI[1]/teiHeader[1]/fileDesc[1]/editionStmt[1]/p[1]::0 /TEI[1]/teiHeader[1]/fileDesc[1]/editionStmt[1]/p[1]::26" xml:id="uid-7b186264-f5ce-4bb2-b060-e3eab90db94f">
      ...
    </annotation>
  </listAnnotation>
</standOff>
```

### DIFF
--- a/src/apps/annotation-text/AnnotatedText/AnnotatedTEI/useEmbeddedAnnotations.ts
+++ b/src/apps/annotation-text/AnnotatedText/AnnotatedTEI/useEmbeddedAnnotations.ts
@@ -132,12 +132,9 @@ export const useEmbeddedTEIAnnotations = (xml?: string) => {
             } as Note)
         );
 
-        const tags: string[] = Array.from(el.querySelectorAll('rs[ana]'))
-          .reduce<string[]>(
-            (all, el) => [...all, ...el.getAttribute('ana')!.split(' ').map((tag) => decodeURI(tag))],
-            []
-          )
-          .map(resolveTag);
+        const tags: string[] = el.getAttribute('ana')
+          ? el.getAttribute('ana')!.split(' ').map((tag) => decodeURI(tag)).map(resolveTag)
+          : [];
 
         const created = changes.find((c) => c.status === 'created');
         const updated = changes.find((c) => c.status === 'modified');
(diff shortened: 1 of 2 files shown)


NickLaiacona (approved):
👍

src/apps/annotation-text/AnnotatedText/AnnotatedTEI/useEmbeddedAnnotations.ts
```diff
@@ -132,12 +132,9 @@ export const useEmbeddedTEIAnnotations = (xml?: string) => {
             } as Note)
         );
 
-        const tags: string[] = Array.from(el.querySelectorAll('rs[ana]'))
-          .reduce<string[]>(
-            (all, el) => [...all, ...el.getAttribute('ana')!.split(' ')],
-            []
-          )
-          .map(resolveTag);
+        const tags: string[] = el.getAttribute('ana')
```
blms, line 135: I wonder if we need to support the old format here as well?

blms (approved):
Looks good, just one question, and we should probably await Rainer's approval first before merging.

src/apps/annotation-text/AnnotatedText/AnnotatedTEI/useEmbeddedAnnotations.ts
ajolipa: Yeah I was wondering about that too. I didn't want to over-complicate the code with making sure to still support the previous format if it's not necessary, but definitely if we expect people to be importing documents that have been exported previously then yeah, we probably should support that.

blms: I could definitely see that happening, if it's not too much trouble to add the support.

rsimon:
Looks good to me! TBH I no longer know what the reasoning behind the `rs` tag was in the first place. I believe it was based on a recommendation (*may* have been by Hugh Cayless - but I frankly no longer know) we got in the "old" Recogito, which used that approach as far as I remember.

rsimon (approved):

NickLaiacona:
@ajolipa ready to merge this one? looks like it also might have a rebase issue

Issue #496, opened 2026-05-04
Email and trigger dependency updates


blms:
> ### Unaddressed - So Far
> 
> * [x] `@1password/sdk` - probably safe to upgrade, but I don't know how to test this locally (used with Trigger.dev)
> * [x] `@react-email` - this might be a bigger chunk, since there was a major version upgrade
> * [x] `nodemailer` - major version upgrade
> 
> ### Note on Security Vulnerabilities
> Some package dependencies currently report security vulnerabilities we should look into:
> 
> * [ ] `@trigger.dev/build` - latest version is vulnerable
> * [x] `nodemailer` - upgrading will probably fix this issue

_Originally posted by @rsimon in https://github.com/recogito/recogito-client/issues/495_

PR #498, opened 2026-05-05
Nodemailer, react-email, 1password SDK dependency upgrades (#496)

blms:
### In this PR

Per #496:
- Upgrade `nodemailer` and 1password SDK (non-breaking for our usage)
- Upgrade `react-email` (only breaking change was the package name)

Unrelated:
- Add a tiny bit of info about Trigger.dev to the readme

### Notes

- I tested invite emails with Ethereal Email, and ran trigger tasks to check the 1pw changes, and all worked fine.
- I'm thinking it's best not to downgrade Trigger.dev versions, and instead wait for the package maintainers to release a new version with the unsafe `systeminformation` and `cookie` versions upgraded. They appear to make a lot of changes between minor releases, so I'm hesitant to downgrade by 5 of them—especially since we are running built versions of the Trigger code in production already. But let me know if you think it's worth a try anyway, and I'm happy to do so!


rsimon (approved):
Looks good!

Issue #497, opened 2026-05-05
Remaining smaller dependency updates


rsimon:
- [x] dotenv
- [x] react-dropzone
- [x] i18next + related packages
- [x] eslint (+ globals)

rsimon:
A note on `react-dropzone`: it looks like this is only really used in one place (the `<InviteListOfUsersDialog>` for uploading a list of project collaborators as CSV).

In the remaining other places (Collection browser, Project documents view, Dashboard project import), it's used only as helper for a conventional upload (click button -> opens file dialog), but never for drag-and-drop. So it might be overkill in these places and something to streamline at a later point.

PR #499, opened 2026-05-06
More Dependency Upgrades

rsimon:
## In this PR

More dependency upgrades, per #497:
- **Updated ESLint to the latest version**
  - Upgraded ESLint config
  - Added a `script` entry to run lint via `npm run lint`
- **Updated and unpinned a few more patch release versions** – Phosphor Icons, Supabase SDK, Uppy, Astro, dotenv, react-textarea-autosize
- **Upgraded i18next**
  - I tested internationalization on server- and client-rendered pages and all worked fine. 
  - When hitting the root route of the app, auto-detection of the browser locale failed! (All browsers always got redirected to the EN version.) I added a fix for this.
- **Upgraded react-dropzone** – I tested click upload (Documents view) and drag-and-drop upload (batch-invite users - the only place we're actually using dnd upload)


src/apps/project-home/ProjectHome.css
```diff
@@ -70,85 +70,4 @@ body {
...
-  margin: 0;
-}
-
-.dropzone-hint-popup p {
-  color: var(--font-semi);
-  font-size: var(--font-tiny);
-  margin: 0;
-  letter-spacing: 0.01ch;
-}
+}
```
rsimon, line 73: Removed some styles that weren't actually used anymore on the page.

src/pages/index.astro
```diff
@@ -1,12 +1,8 @@
 ---
 import { getMyProfile } from '@backend/crud';
 import { createSupabaseServerClient } from '@backend/supabaseServerClient';
-import { i18n } from 'astro:config/server';
 
-const supported = i18n?.locales || ['en', 'de'];
-const preferred = Astro.preferredLocale;
-
-const lang = Astro.currentLocale || (preferred && supported.includes(preferred) ? preferred : 'en');
+const lang = Astro.preferredLocale ?? 'en';
```
rsimon, line 5: As far as I could figure out, `Astro.currentLocale` always gets set to the default language (en) if the user isn't on a page with a `/{lang}/` path yet. Therefore, the `lang` parameter always go set to `en` here, regardless of browser locale.

blms: Ah, good catch! Thanks for the fix.

blms (approved):
Looks great! Thanks for the i18n fix and the lint script. Astro config looks simpler too.

PR #501, opened 2026-05-14
Resolve some linter errors and warnings

blms:
### In this PR

Resolving some linter errors/warnings, in particular:
- `no-undef`
- `react/display-name` rule disabled despite not being configured
- `react-hooks/immutability`
- `react-hooks/use-memo`
- `preserve-caught-error`
- Some `exhaustive-deps`

We should probably smoke test at least some of these.

I was going through these by running:
```sh
npx eslint . --rule 'react-hooks/set-state-in-effect: off' --rule 'react-hooks/refs: off' --rule 'react-hooks/static-components: off'
```

since those three rules seemed really tricky to deal with. At this point, all that shows up with that command are about 80 `exhaustive-deps` warnings, which can be annoying because you have to test each case for accidental infinite re-renders. There's probably a good chunk of them we'll just want to disable.

### DIFF
--- a/src/components/AnnotationDesktop/ColorCoding/colorCodings/useColorByFirstTag.ts
+++ b/src/components/AnnotationDesktop/ColorCoding/colorCodings/useColorByFirstTag.ts
@@ -17,16 +17,20 @@ export const useColorByFirstTag = (vocabulary: VocabularyTerm[] = []): ColorCodi
 
   const tags = useMemo(() => enumerateTags(annotations), [annotations]);
 
+  const vocabularyJSON = JSON.stringify(vocabulary);
+
   const getColor = useMemo(() => {
     const palette = createPalette(PALETTE);
 
-    const getColor = (tag: string)  => {
-      const preset = vocabulary.find(t => t.label === tag)?.color as Color;
+    const parsedVocabulary = JSON.parse(vocabularyJSON) as VocabularyTerm[];
+
+    const getColorFn = (tag: string)  => {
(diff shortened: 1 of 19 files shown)


src/components/AnnotationDesktop/ColorCoding/colorCodings/useColorByFirstTag.ts
```diff
@@ -17,16 +17,20 @@ export const useColorByFirstTag = (vocabulary: VocabularyTerm[] = []): ColorCodi
 
   const tags = useMemo(() => enumerateTags(annotations), [annotations]);
 
+  const vocabularyJSON = JSON.stringify(vocabulary);
```
rsimon, line 20: This is probably worth looking into more deeply at some point. I wonder why this is even needed to be checked by value rather than object reference.

(FWIW: it looks like vocabulary should be stable - it's created [here](https://github.com/recogito/recogito-client/blob/develop/src/backend/hooks/useTagVocabulary.ts).) But for now this definitely removes the linter warnings.

rsimon (approved):
Thanks - these all look good & safe!

rsimon:
@blms P.S. I tested removing the `JSON.stringify` "by value" dependency in both text and image annotation mode, and there seem to be no negative effects. Unless I missed something, this may have just been some over-defensiveness on my part, or a legacy issue. I can send a separate PR for that later if you want!

blms:
That would be great, thank you @rsimon !

PR #502, opened 2026-05-16
Minor Fixes: Linter Warning, Tag Autosuggest Popover

rsimon:
## In this PR

As discussed in https://github.com/recogito/recogito-client/pull/501#pullrequestreview-4295731597, this PR removes the legacy hack in the "first tag" color coding hook.

This PR also fixes a bug I noticed while testing: the tag autocomplete dropdown was rendered as a child to the annotation card, which had `overflow-y` set, creating an odd broken layout:

Bildschirmfoto 2026-05-16 um 09 36 29

I'm pretty sure the extra overflow rule was added after we realized that some users were writing very long comments, and that would create annotation cards that exceed the screen height. The fix, fortunately, was easy. The dropdown is now in a portal.


blms (approved):
Looks good, and thanks for catching/fixing the bug!

PR #503, opened 2026-05-21
Upgraded to Text Annotator 4.1

rsimon:
## In this PR

This PR upgrades to Text Annotator v4.1, which includes significant UI performance improvements, in particular for TEI/XML annotation. 

Details on Text Annotator changes:
- https://github.com/recogito/text-annotator-js/pull/261
- https://github.com/recogito/text-annotator-js/pull/262

Note that [PR 262](https://github.com/recogito/text-annotator-js/pull/262) dropped the `start` and `end` character offset fields from TEI selectors. Instead, a new `position` field was introduced which allows for fast sorting of annotations by document order.


blms (approved):
This is great!

Are there any breaking changes here from an end-user perspective, or does the new sorting logic take care of any changes in the library?

rsimon:
No changes in the UX. The new "position" field works just like a char offset, and the new sort logic in this PR changes the sort field depending on the document type as needed. (We already had different sorting for PDF and other text before, so that's a minimal change.)

Also: XPaths expressions are considered the only source of truth, and positions always get re-generated from the XPaths on load (even if the input annotation already happens to have a position). Therefore, everything is fully backwards compatible, and should cause no problems with any of the existing annotations already in the system.

PR #504, opened 2026-05-27
Fix infinite re-render in invite user dialog

blms:
Quick bug fix for something I found working locally, introduced by #501 as the `formik` dependency caused infinite re-renders.


rsimon:
Yikes - yes, the dangers of the useEffect dependency array... good solution!

rsimon (approved):

PR #505, opened 2026-05-28
FIX: browser based routing for translations

jsmueller7:
### Problem

This issue started from client `v.1.8.0`. Earlier versions didn't show this behavior: When users navigate to `/`, the application always redirected to `/en/` regardless of their browser's language preference, even when German (`de`) was configured as a supported locale.

### Root Cause

Two compounding issues:

**1. Astro i18n middleware intercepting `/` before the page ran**

`astro.config.node.mjs` had `i18n.routing` configured with `strategy: "pathname-prefix-always"` and no explicit `redirectToDefaultLocale` setting, which defaults to `true`. This caused Astro's built-in i18n middleware to intercept every request to `/` and immediately issue a `302` redirect to `/{defaultLocale}/` (i.e. `/en/`) — before `src/pages/index.astro` was ever executed.

**2. `index.astro` not reading `Accept-Language`**

The original `index.astro` (i.e. `v1.7.0` and lower) used `Astro.preferredLocale` / `Astro.currentLocale` for locale detection, which only inspects the URL path — not the browser's `Accept-Language` header. Even if the middleware had been disabled, the page would still have fallen back to the default locale for all requests to `/`.

### Changes

**`src/pages/index.astro`**

Replaced Astro's path-based locale detection with explicit `Accept-Language` header parsing using the `accept-language-parser` package:

```js
import parser from 'accept-language-parser';

const supported = i18n?.locales || ['en', 'de'];
const defaultLocale = i18n?.defaultLocale ?? 'en';
const accepts = Astro.request.headers.get('accept-language');
const lang = accepts
  ? parser.pick(supported, accepts) ?? defaultLocale
  : defaultLocale;
```

**`astro.config.node.mjs`** (relevant for a containerized setup)

Added `redirectToDefaultLocale: false` to the `i18n.routing` block to prevent Astro's middleware from short-circuiting requests to `/` before `index.astro` runs:

```js
i18n: {
  defaultLocale: 'en',
  locales: ['en', 'de'],
  routing: {
    redirectToDefaultLocale: false,  // allow index.astro to handle /
    // ... existing settings unchanged
  }
}
```

### Behaviour after fix

| Browser language | Redirect target |
|---|---|
| `de`, `de-DE`, `de-AT`, etc. | `/de/…` |
| `en`, `en-US`, `en-GB`, etc. | `/en/…` |
| Unsupported / no header | `/en/…` (default fallback) |

Existing behaviour for all other routes (`/en/*`, `/de/*`) is **unchanged**.

### Dependencies

`accept-language-parser` — already present in `package.json`.

### DIFF
--- a/astro.config.node.mjs
+++ b/astro.config.node.mjs
@@ -19,6 +19,7 @@ export default defineConfig({
     locales: ['en', 'de'],
     defaultLocale: 'en',
     routing: {
+      redirectToDefaultLocale: false,
       prefixDefaultLocale: true,
     },
   },
--- a/src/pages/index.astro
+++ b/src/pages/index.astro
@@ -1,8 +1,14 @@
 ---
 import { getMyProfile } from '@backend/crud';
 import { createSupabaseServerClient } from '@backend/supabaseServerClient';
+import { i18n } from 'astro:config/server';
+import parser from 'accept-language-parser';
 
-const lang = Astro.preferredLocale ?? 'en';
+const supported = (i18n?.locales as string[]) || ['en', 'de'];
+const defaultLocale = i18n?.defaultLocale || 'en';
+
+const accepts = Astro.request.headers.get('accept-language');
+const lang = accepts ? (parser.pick(supported, accepts) ?? defaultLocale) : defaultLocale;
 
 const supabase = await createSupabaseServerClient(Astro.request, Astro.cookies);
 


rsimon:
Hi @jsmueller7, 

Did you test the error behavior against the latest state of the `develop` branch? I noticed the same issue a few weeks ago, and then made the following fix (as part of PR #499):

https://github.com/recogito/recogito-client/pull/499/changes/f8cf036919e7f9db209ebeea62b40008cb0776b5

TL;DR: `preferredLocale` should do the trick once `navigator` is added to the i18n detection options (which is what the above commit does.) This seemed to solve the issue.

The commit is already merged into `develop`. But it is not yet part of the latest official release, which is why the latest release would still have the issue.

jsmueller7:
Hi @rsimon: No, sorry. I didn't check against the current state (mainly because I don't have a develop pipeline).

Theres another minor i18n issue. The container throws this message:
> eact-i18next:: useTranslation: You will need to pass in an i18next instance by using initReactI18next { code: 'NO_I18NEXT_INSTANCE' }

This seems to be uncritical, but while looking into the translations issue I looked into it. In my own limited words: The source is the BrandFooter.tsx that calles `useTranslation` without an i18n client instance. When patching the BrandFooter.tsx the issue is gone, but there's a kind of "cold start issue": When I open the recogito site for the first time it takes 3 seconds extra to initialize.

Is it woth that I look further into it?

jsmueller7:
Hi @rsimon: I just created a build from the develop branch and I can confirm that the routing is fixed there. But:

- The fix might be slowing down (I can't say for sure, because I use the mobile network today. So my internet is overall slow)
- The console prints these warnings: 

> [WARN[] Internal Warning: route cache overwritten. (/[lang[]/sign-in_src/pages/[lang]/sign-in.astro)

From an operator perspective the fix in this PR works better: Quicker, no warnings. (Note: I didn't patch `astro.config.mjs`, as I am using the node server, so if you want to test this yourself, you might need to also add the line `redirectToDefaultLocale: false,` in the `astro.config.mjs` too).

You should do what you think is best; it's your field of expertise. I just threw an extensive amount of time, tokens and tests against this untill I found this to be the best fix.

rsimon:
@jsmueller7 thanks, I could reproduce the warning and can confirm your fix prevents it. 

I dug into the matter a bit more though, and it appears that (although I'm not 100% sure) your code effectively performs the same functionality as calling `preferredLocale`. However: `preferredLocale` ran into a known bug in Astro – which was then fixed very recently (May 26, Astro 6.3.8): https://astro-changelog.netlify.app/releases/re_kwdofl76q84tox3c/

I can also confirm on my end that upgrading Astro fixes that warning when using the old code. (I tested this specifically also with the Node self-hosting build.)

I didn't see any difference in speed between both approaches TBH (when looking at the Chrome dev tools load time measurement).

My personal tendency, therefore, would still lightly be towards sticking with the current one-line `const lang = Astro.preferredLocale ?? 'en';` version. But I honestly don't have a strong opinion, especially since this wouldn't a big change either way.

@blms what do you think? Could you also take a quick look and check if you seen any pros or cons to either approach, especially concerning performance?

jsmueller7:
@rsimon The speed issue then must have resulted from me using the mobile newtwork. I am fine with any final solution when it doesn't create other issues 🙂

blms:
@rsimon @jsmueller7 Thanks both. Testing locally I'm not seeing a performance difference, and I prefer the simpler approach to fixing the warning (astro upgrade)

jsmueller7:
Allright. I'll close the PR as it became obsolete.

PR #518, opened 2026-06-29
Bump SDK to 0.9.0, deduplicate from new SDK code, and bump some other packages

blms:
### In this PR

- Bump SDK to 0.9.0
- Remove all of the migrated client code and update imports to use SDK exports instead
- Bump minor versions of annotorious/annotator packages
- Fix some phosphor icon imports to resolve deprecation warnings by using the new naming style (I've just been gradually doing this across the repo)


rsimon (approved):
Looks good, thanks! 

Is the Filter setting working correctly at runtime? (Importing contexts from external libraries always seems tricky. But the way the React dependency is set up in the SDK should make it work correctly 🤞)

blms:
Good call, it seems to work fine from local testing at least.

PR #521, opened 2026-07-02
Improve plugin installation workflow with env var

blms:
### In this PR

- Add `INSTALLED_PLUGINS` env var and parse it to populate astro config plugin integrations
- Add `npm install` step to the `Dockerfile` so that self-hosted installations get them installed when running install/upgrade scripts

### Notes

I realized while working on the self-hosting documentation that there's not really a clean way to integrate plugins while self-hosting, which has until now required maintaining a separate `astro.config.mjs` with the plugins listed in `integrations`. In practice, that means keeping a fork or modified clone of this repo, with the plugins imported in that file as well as installed via NPM.

I also have an existing [easy-upgrade shell script](https://github.com/recogito/recogito-studio/blob/main/upgrade.sh) in the self-hosting repo that is meant to upgrade everything end-to-end, but I never included logic for adding plugins with it, and therefore it uninstalls them on every upgrade.

We handled this in `recogito-cloud` by maintaining [separate copies of the Astro configs](https://github.com/performant-software/recogito-cloud/tree/develop/plugin-configs), but those tended to drift and added a maintenance burden. We already have a `PLUGINS` env var there to configure plugins per-tenant, so I took a cue from that, and from Django's `INSTALLED_APPS` setting, and added a new `INSTALLED_PLUGINS` env var to this repo.

Not sure whether to consider this a breaking change or not; technically anyone doing it the old way will have no breakage (they already maintain a separate copy of the config if they are using plugins), and omitting the env var on a fresh install just means you get no plugins.

### LLM Note

Claude wrote the parser script and dockerfile command (though they're pretty simple in retrospect).

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -35,6 +35,17 @@ PUBLIC_ENABLE_USER_INVITE=TRUE    # When 'TRUE' Org Admins can invite email/pass
 # For Node builds, set to the public web address of your instance
 SITE_URL="http://localhost:4321"
 
+# Plugins to install and register, as a comma- or space-separated list of npm
+# package names. The packages must also be installed (npm install <name>).
+# Available Recogito Studio plugins:
+#   @recogito/plugin-geotagging             GeoTagger
+#   @recogito/plugin-reconciliation-service Reconciliation Service API
+#   @recogito/plugin-ner                    Named Entity Recognition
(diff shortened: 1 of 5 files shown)


rsimon (approved):
That's a pretty ingenuous idea - I love it!

PR #522, opened 2026-07-06
Handle no component in doc-actions extension point

blms:
### In this PR

- Minor CSS change that goes along with https://github.com/recogito/metaphor-annotation-plugin/pull/5, to hide double separators when there is no mounted (lazy-loaded) component


rsimon (approved):

rsimon (approved):
Makes sense!

PR #525, opened 2026-07-28
`develop` into `main` for 1.10.0

blms:
leaving as draft while I figure out what needs to be tested, documented, etc

- [x] trigger deploy (now part of deploy workflows)
- [ ] expose recogito_etl schema


NickLaiacona:
> ### testing notes
> on [staging](https://recogito-staging.covecollective.org/):
> 
> * [x]  Add a Getty IIIF image to a project and annotate it
> * [x]  To test annotator upgrades: Create and edit annotations in IIIF, PDF, TEI, and plaintext docs
> * [x]  To test annotator upgrades: View and comment on existing annotations in IIIF, PDF, TEI, and plaintext docs
> * [x]  Export a project (ideally one without large IIIF images, see below Trigger note) by using the “three vertical dots” menu on a project (visible from the project dashboard) as an org admin, and clicking “Export project”
>   
>   * [x]  You’ll have to manually refresh the jobs list page to see updated progress, and to download the export once it has completed. You’ll also be able to see if a job errored here (not real-time)

Regarding the project import, I imported the ZIP file from the export, it says it completed successfully but I still only have one project, should this create a duplicate project?

blms:
> Regarding the project import, I imported the ZIP file from the export, it says it completed successfully but I still only have one project, should this create a duplicate project?

@NickLaiacona Hmm, it should. Was it a public or shared project, or one that appeared in your own "my projects" dashboard? The duplicate should appear in the same place of those three that the original project was. If it's not showing up anywhere, let me know the original project ID and I'll take a look

NickLaiacona:
It was in "My Projects" and there doesn't seem to be another copy there. Here's the original project: https://recogito-staging.covecollective.org/en/projects/3cb0b400-f941-4aa3-9ff3-ba7b13198bb5 . The import was at 12:37 Aug 06 .

NickLaiacona:
- [x] Invite new users, both from the instance root and from a project, and ensure the emails are sent and invites work (can just be nick+whatever@performant emails)
- [ ] To test changes in the SDK: 
   - [ ] On a document with multiple annotations, filter by contributor/tag/visibility, color-code by tag/layer/privacy, and use a tag autosuggest field


OK, a couple of questions about how annotation works these days. My setup is that I am logged in with two different accounts on two different browsers (Edge and DuckDuckGo) into the same assigned document with all layers visible. I have user A who created the assignment and user B that is a student. 

User A is annotating in the base layer and user B is annotating in the assignment layer. For the first several annotations, user B's name is displayed on his annotations. But now, they are coming across as anonymous. Not sure why. This seems to also be happening for user B's annotations now from user A's perspective.


Screenshot 2026-08-07 114348
Screenshot 2026-08-07 114309

NickLaiacona:
Regarding the plugins - I'm presently an Org Admin and I don't see a way to activate plugins. Do I need a higher level of permission to do this?

blms:
@NickLaiacona looking into the anonymous thing now!

As for plugins, you should be able to as an org admin. Go to a project that you own or admin, and from the project's page Settings --> Plugins --> Browse Available Plugins

NickLaiacona:
ah ok, I see it! However, I get this:

Screenshot 2026-08-07 120001

blms:
@NickLaiacona Hmm ok, I thought I'd fixed that, but looks like maybe not. Looking further into it now.

For the anonymous thing, could you try refreshing one or both browsers where you see Anonymous? Does that show the correct user name?

It might be something to do with backgrounding applications interacting with supabase presence. It might need better fallback behavior when presence falls off or fails to make it to the client.

blms:
@NickLaiacona Redeployed and it looks like the plugin gallery is loading again, let me know if it's still empty for you after a hard refresh

NickLaiacona:
@blms , ok - I see the plugins now and the annotations from before now are attributed to their owners. I'm also seeing the annotation focus of the other user, which I wasn't seeing earlier.

NickLaiacona:
OK, testing out the plugins:
### TEI Inliner Plugin
 
I exported the "Little Princess" document I was annotating earlier with the option of "Export TEI: Convert Annotations to Inline"->"Recogito Annotations Only". When I examine the TEI, I don't see any inline annotations at locations where I expect them. Also don't see any standoff.

Unrelated, but it seems that the special entities are not being escaped in the inline text, which technically makes this file invalid XML, see screenshot.

Screenshot 2026-08-07 125111

### NER plugin

I ran the Stanford NLP against my test document giving the output filename of "Princess NER" or something like that. I haven't heard from it since. I don't see a job queued in the Jobs table or anything like that.

### Reconciliation Service API

I found a name of a place in the document, marked it and looked it up. This worked fine in the sidebar view of the annotation editor. When it was rendered as a floating dialog box, I got this weird looking invisible frame with a horizontal scrollbar. See screenshot.

Screenshot 2026-08-07 124353

Screenshot 2026-08-07 124413


### Untested plugins:

I don't know the first thing about these other plugins, so not sure I can accurately test them:

* Recogito Revisions Plugin
* Sandcastle3D Export
* GeoTagger

blms:
Thanks very much @NickLaiacona! It looks like we've got some work cut out for us on the plugins before shipping. Luckily the bochum budget landed so we can get on these ASAP early next week.

blms:
> I exported the "Little Princess" document I was annotating earlier with the option of "Export TEI: Convert Annotations to Inline"->"Recogito Annotations Only". When I examine the TEI, I don't see any inline annotations at locations where I expect them. Also don't see any standoff.

Apparently right now this only works with annotations where there is a tag on the annotation, written exactly as `placeName` or `persName`. All other annotations are silently dropped.

I don't know what the intention was overall, but that does seem to [match the readme for the plugin](https://github.com/recogito/plugin-tei-inliner#features). Perhaps this needs more documentation in RS itself to clarify the limitations?

> Unrelated, but it seems that the special entities are not being escaped in the inline text, which technically makes this file invalid XML, see screenshot.

Weird—I'm not seeing that, in the same place I see `&amp`:
Screenshot 2026-08-10 at 4 52 36 PM
How are you examining it? Could it have been somehow transformed by the time you're reading it?

blms:
> I ran the Stanford NLP against my test document giving the output filename of "Princess NER" or something like that. I haven't heard from it since. I don't see a job queued in the Jobs table or anything like that.

Looks like this one completed: https://recogito-staging.covecollective.org/en/annotate/3ec8cec2-3519-4176-9673-25cc2fcc63b5/4caa8f0b-9d8f-41a0-b5a0-cf8d72237a33
In this project: https://recogito-staging.covecollective.org/en/projects/3cb0b400-f941-4aa3-9ff3-ba7b13198bb5

However, I do seem to be unable to load any of the actual annotations here, I see Anonymous with the "typing" animation, so I'm going to investigate that further… edit: working on it in https://github.com/recogito/plugin-ner/pull/5

FWIW: It doesn't show up in the Jobs table at this point—that's only available to Org Admins and only stores info about the import/export jobs. I don't think we have a way of tracking NER tasks at the moment outside of trigger, but that's a pre-existing issue.

blms:
- Reconciliation service fix in this PR: recogito/plugin-reconciliation-service#5
- I also tested both revisions and the geotagger plugin and both seem to be working as expected.

The only remaining one is Sandcastle3D export, which I may need some help with as I'm not sure what the expected output needs to be shaped like. Though, from a quick test, the menu option and export endpoint seem to work:
   Screenshot 2026-08-11 at 5 06 39 PM

rsimon:
@blms - I took a quick look at the sandcastle plugin:

- The plugin was kept (intentionally) as simple as possible. Duke University's main request was that they didn't want to deal with the stringified W3C annotation selectors, but instead use our structured JSON format (for boxes, polygons, etc.) directy. So the plugin simply serializes our internal annotations as-is, without applying the W3C target crosswalk. The only thing the Sandcastle plugin really does is to crosswalk the annotation bodies from the (proprietary, JSON-based) Quill format to HTML.
- I updated the SDK dependency and ran a quick test with a local install, against the latest `develop` branch of `recogito-client`. All seemed to work well - see [PR description](https://github.com/recogito/plugin-sandcastle-export/pull/1).

blms:
Thanks @rsimon! Could we mint new versions of all the changed plugins once you've reviewed/merged the remaining stuff on NER? (I guess the NER plugin will need to be updated to consume the new standoff converter.) That should be the last bit needed for this prod release 🤞

rsimon:
New releases are out for:
- [x] [plugin-sandcastle-export](https://www.npmjs.com/package/@recogito/plugin-sandcastle-export) (0.2.0)
- [x] [plugin-reconciliation-service](https://www.npmjs.com/package/@recogito/plugin-reconciliation-service) (0.10.0)
- [x] [standoff-converter](https://www.npmjs.com/package/@recogito/standoff-converter) (0.2.0)
- [x] [plugin-ner](https://www.npmjs.com/package/@recogito/plugin-ner/v/0.5.1) (0.5.1)
- [x] [plugin-revisions](https://www.npmjs.com/package/@recogito/plugin-revisions/v/0.2.0) (0.2.0)
- [x] [plugin-geotagging](https://www.npmjs.com/package/@recogito/plugin-geotagging) (0.6.0)
- [x] [plugin-tei-inliner](https://www.npmjs.com/package/@recogito/plugin-tei-inliner/v/0.2.0) (0.2.0)

